PR #1, opened 2019-07-09
Stardog transaction support

DarthStem:
Changes needed to support transactions with the Stardog Store.  

* Requires Stardog v6.1.1 or later due to some bugs I discovered when querying into an open transaction with reasoning enabled.  (see https://www.stardog.com/docs/release-notes/#_6_1_1_release_2019_02_06 #6794, #6796)
* Requires dotNetRdf v2.2.0 in order to support querying into an open transaction with reasoning (see https://github.com/dotnetrdf/dotnetrdf/pull/226)


faubulous:
Merged into branch release/1.0

PR #4, opened 2019-10-21
Update StardogTransaction.cs

DarthStem:
I have a use case where I require access to the StardogConnector while a transaction is executing.  Is it possible to change the access of the underlying connector to public (ReadOnly)?


meberl:
Merged into develop

PR #5, opened 2020-03-04
Test for Resource with multiple types issue

joephayes:
If a resource has multiple RDF types (that is, John Lennon is both a solo artist and a songwriter), it is not possible to retrieve it using those multiple types - if we retrieve John Lennon as a songwriter and then try to retrieve him as a solo artists we will get back the songwriter.


joephayes:
@faubulous @meberl

PR #37, opened 2026-08-19
Layered read views: baseline + additions − removals

meberl:
## What

A `ModelGroup` can only union: its dataset clause is one `FROM` per member, and SPARQL has no inverse of that. This adds `ILayeredModel` — a read view over a **baseline** graph plus a pending change held as an **additions** and a **removals** graph — so everything reading through it sees `(baseline − removals) ∪ additions` while the baseline itself stays untouched until the change is accepted or discarded. A git working tree, expressed in RDF.

Mapped-object reads are the point: `GetResource`, `GetResource<T>`, `GetResources<T>` and `ContainsResource` honour the overlay, so code written against `IModel` needs no change.

```csharp
var view = store.CreateLayeredModel(baseline, additions, removals);

// stage a change by writing to the layer models, which are ordinary IModels
// then read through the view as if it had been applied
var person = view.GetResource<Person>(uri);          // sees the staged value
view.AsQueryable<Person>().Where(p => p.Name == "…"); // and so does selection
```

A triple in **both** additions and removals is **visible** — additions win. That is what makes an ordinary value edit (remove the old, add the new) behave the way callers expect.

## Design

**Not an `IModelGroup`.** A group is an `ISet<IModel>` of interchangeable members; these three graphs have distinct roles, so set membership means nothing for them. Keeping the types apart also stops the existing union-only paths — the `is IModelGroup` branches in the `SparqlQuery` model setter and `GenerateDatasetClause`, and GraphDB's inference path — from mistaking a view for a plain group. **The `IModelGroup` contract is untouched and no existing consumer changes behaviour.**

**Created via a `StoreExtensions` method, not a new `IStore` member**, so no backend has to change, and the empty-group bug that three stores carry in `CreateModelGroup(params IModel[])` is not reproduced. The constructor is `internal`: all three graphs must live in one store, because the overlay is a single query over a single dataset. Composing one across stores was measured and **fails silently in both directions** — returning triples staged for removal, or dropping the baseline entirely — so routing construction through the factory makes that unrepresentable rather than merely discouraged.

**Read-only.** Stage a change by writing to `Additions`/`Removals`. Routing writes automatically would mean splitting the `WITH <g> DELETE {…} INSERT {…}` delta from `StoreBase.TryBuildDeltaUpdate` so its DELETE half becomes an INSERT into the removals graph — a new write path in every store. Deferred.

**`IModelGroup.DefaultModel` turns out to be dead** — declared, auto-implemented, never read or assigned anywhere. Recorded in ADR-0019; removing it is a breaking public-interface change and left for its own PR.

## Three emission rules that are requirements, not optimisations

Subtraction has to live in the graph pattern, so `LayeredModelSparql` is the single source of truth. Each rule below was measured, and getting one wrong is either a silent wrong answer or four orders of magnitude:

1. **`MINUS` where the pattern binds a variable, `FILTER NOT EXISTS` where it is fully ground.** `MINUS` is far faster — an anti-join instead of a per-row probe, 18× → 1.5× on an unselective scan — but it removes nothing when the two sides share no variables, and a ground pattern has none. There it **silently fails to subtract**. Verified: the in-memory store and GraphDB are spec-correct here; Virtuoso is the deviation.
2. **Bind a known subject with `VALUES` before the overlay**, never as a trailing `FILTER`. Neither the in-memory engine nor GraphDB pushes a filter into a `UNION` containing an anti-join — 15× and 38× against ~1.0×.
3. **Selective patterns before the wildcard `?s ?p ?o`.** An engine that cannot reorder joins across the overlay's `UNION` evaluates them as written, so a leading wildcard materialises the whole effective graph first: **61,000×** in-memory. `BuildResourceQuery` previously emitted the wildcard first — hence the change there. Pattern order inside a BGP is semantically irrelevant, so plain models are unaffected.

## Query coverage

**LINQ is honoured natively** because Trinity owns its AST: only `TriplePattern` resolves against data, so wrapping one `case` arm of `SparqlQueryWriter` is exhaustive by construction, and the existing `default: throw` keeps a future node type from leaking.

**Caller SPARQL is rewritten** by `OverlayQueryRewriter`, working on the parse tree rather than the text — the abbreviations that defeat a token-level rewrite (`;`, `,`, `[ ]`) are already expanded by the time the parser is done. It is a whitelist; anything unrecognised throws with a reason.

| Refused | Why |
|---|---|
| property paths | walks intermediate nodes that must themselves resolve against the effective graph; `+`/`*` have no fixed expansion |
| `GRAPH` blocks | a view is built from three graphs; naming one reads past the overlay, and `GRAPH ?g` exposes the removals graph as data |
| `SERVICE` | the remote endpoint knows nothing of the overlay |
| `CONSTRUCT` / `DESCRIBE` | a template describes triples to build, not match; `DESCRIBE` has no pattern to rewrite |
| `FILTER EXISTS` / `NOT EXISTS` | an `ExistsFunction` inside the filter expression, whose nested pattern would read the baseline directly |
| a caller `FROM` / `FROM NAMED` | the view defines the dataset |
| `inferenceEnabled: true` | every store's inference path defeats the overlay, and entailment over a subtracted graph is undefined |

Refusing matters because running such a query unchanged is wrong **two different ways**, neither of which announces itself: a default-graph pattern returns *nothing* (the view names its graphs with `FROM NAMED`, leaving the default graph empty), while a pattern naming a graph returns triples staged for removal.

## An upstream serialisation defect, contained

dotNetRDF drops the parentheses around a negation's operand: `!(?r < 3)` → `!?r < 3`, i.e. `(!?r) < 3` — **and it still parses**. `SparqlFormatter` has the same defect. Parsing is fine; only writing out is wrong.

`SparqlExpressionWriter` therefore serialises expressions itself, parenthesising every operator application so correctness needs no precedence table. **22/22 faithful against dotNetRDF's 17/22.** `HAVING` and projected expressions come from dotNetRDF's serialisation of the query head and cannot be repaired here, only detected — so a negation in either is refused with a suggested rephrasing.

**Every rewrite is re-parsed and compared structurally** against the original — query form, `LIMIT`, `OFFSET`, projected variables, `HAVING`, projections, and every filter and `BIND` expression. That check found two defects during development that no test of the output would have: clauses held in `UnplacedFilters`/`UnplacedAssignments` were being **dropped from the rewrite entirely**, and one filter reachable through both `Filter` and `UnplacedFilters` was double-counted.

## Performance

Layered ÷ plain at **1,000,000** baseline triples, ~100 staged additions and ~100 staged removals:

| Read shape | in-memory | Virtuoso | GraphDB |
|---|---|---|---|
| `GetResource` (subject-bound) | 1.55× | 1.06× | 0.86× |
| LINQ `Where` (selective) | 3.00× | 1.03× | 1.55× |
| `GetResources<T>()` (unselective scan) | 1.50× | 1.05× | 1.14× |

**The cheaper-sounding alternative was measured and rejected.** Server-side `COPY` works on all three stores — correcting an expectation that Virtuoso's `SPARQL { … }` update wrapper would reject it — but costs 3.9–6.0 s and a full duplicate of the baseline *per working copy* at 1M triples, and moves the accept-time cost to an O(baseline) diff instead of reading an O(changes) delta.

**Accepted cost, documented as a follow-up:** a caller query is parsed, walked, re-emitted and re-parsed on *every* execution, because the strict parser rejects Trinity's `@parameter` syntax so the rewrite runs after binding. ADR-0041 records the fix for when it matters — substitute each `@parameter` with a placeholder IRI before parsing, rewrite once, cache, map back — which makes it a per-query-shape cost, needs no parser change, and keeps the verification applicable. Affects caller SPARQL only; mapped reads and LINQ never touch the rewriter.

## Tests

Run against **every backend with a fixture**, because the union half of the overlay is exactly where backends differ. `LayeredModelQueryCorpusTest` is a data-driven corpus of caller SPARQL with expected results plus the forms that must be refused; every case is built so the answer *differs* depending on whether the overlay was applied, so a query ignoring subtraction fails rather than coincidentally agreeing.

| Suite | develop | this branch | Δ | failure set |
|---|---|---|---|---|
| in-memory | 410 P / 3 F / 7 S | **542 P** / 3 F / 7 S | +132 | identical |
| Virtuoso | 103 P / 4 F / 1 S | **175 P** / 4 F / 1 S | +72 | identical |
| GraphDB | 110 P / 4 F / 1 S | **182 P** / 4 F / 1 S | +72 | identical |
| Fuseki | 4 P / 82 F | 4 P / 82 F | +0 | identical |
| Generator / Vocabulary | 23 P / 29 P | 23 P / 29 P | — | clean |

**No new failures on any backend.** I measured each `develop` baseline in a clean worktree under the same conditions and diffed the failure *names*, not just the counts — byte-identical throughout.

⚠️ **Pre-existing failures a reviewer will see, all reproducing on `develop`:**
- **in-memory (3)** — `Equal`, `ResourceConstructorTest`, `EqualsTest`. Appear to be an artefact of running under `DOTNET_ROLL_FORWARD=LatestMajor` on .NET 10 where no net8.0 runtime is installed.
- **Virtuoso / GraphDB (4 each)** — the inferencing tests, consistent with this design refusing inferencing.
- **Fuseki (82)** — the documented 4/86, HTTP 404/405 from the upstream `FusekiConnector` bug (ADR-0036). This branch adds no Fuseki tests; the 79 distinct failure names match `develop` exactly, which was the open question given the `SparqlSerializer` and LINQ translator changes.

Store suites are Docker/Testcontainers and excluded from the default CI job (ADR-0036), so CI will only show the in-memory run. All three were provisioned fresh locally and are reported above.

## Non-goals

Branching, merge, conflict detection, history, nested or stacked overlays, and views spanning more than one store. One baseline, one additions graph, one removals graph, one store.

## Review guide

Three commits, each of which builds (verified in a throwaway worktree):

1. **`1e6a5c9`** product code — `Trinity.csproj` builds alone
2. **`bab3029`** tests — solution builds, suite green
3. **`2a3c80b`** docs — ADR-0041, the ADR-0019 note, index, `CLAUDE.md`

`LayeredModel` and `OverlayQueryRewriter` reference each other, so the product code could not be sliced finer without a non-building intermediate.

**Start with `doc/adr/0041-layered-read-views.md`** — it carries the reasoning, the measured numbers, and the reasons behind each refusal.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,7 +55,7 @@ is netstandard2.0 / net8.0 and builds cross-platform.
 
 ```bash
 dotnet build Semiodesk.Trinity.sln -c Release          # whole solution, SDK-only
-dotnet test Trinity.Tests/Trinity.Tests.csproj         # 413 passed, 7 skipped (quarantined)
+dotnet test Trinity.Tests/Trinity.Tests.csproj         # 542 passed, 7 skipped (quarantined)
 dotnet test tests/Trinity.Generator.Tests/Trinity.Generator.Tests.csproj   # 23 passed
 dotnet test tests/Trinity.Vocabulary.Tests/Trinity.Vocabulary.Tests.csproj # 29 passed
 dotnet pack Trinity/Trinity.csproj -c Release          # -> Semiodesk.Trinity.2.0.0.nupkg
@@ -69,7 +69,7 @@ dotnet pack Trinity/Trinity.csproj -c Release          # -> Semiodesk.Trinity.2.
 - Store integration tests (`tests/Trinity.Tests.*`) **self-provision** their server in Docker via
   Testcontainers on a random host port (ADR-0036): run `dotnet test tests/Trinity.Tests.{Virtuoso,GraphDB,Fuseki}`
   with a Docker daemon running. Excluded from the default CI job (Docker + large images). Current:
-  Virtuoso 103/108 and GraphDB 110/115 pass; Fuseki is 4/86 — a pre-existing dotNetRDF `FusekiConnector`
+  Virtuoso 175/180 and GraphDB 182/187 pass; Fuseki is 4/86 — a pre-existing dotNetRDF `FusekiConnector`
   query-endpoint bug (POSTs `/ds/query`, which the server 404s), unrelated to the container wiring.
   Virtuoso's `Int16Test`/`Uint16Test`/`UintTest` are now `Assert.Inconclusive` in `VirtuosoResourceTest`,
   alongside the pre-existing `Int64Test`/`Uint64Test` overrides for the same phenomenon: Virtuoso widens
@@ -168,6 +168,23 @@ Invariants that surprise newcomers:
   `GetValue`/`SetValue` + `GetTypes()`, now emitted by the source generator (ADR-0013). A
   `[RdfProperty]` on a non-`partial` member is not generated (and no longer woven) — it does nothing.
 - **Models are named graphs; a `ModelGroup` is itself an `IModel`** spanning several (0019).
+- **A `ModelGroup` can only union; subtraction is a separate abstraction** (0041): `ILayeredModel`
+  (`store.CreateLayeredModel(baseline, additions, removals)`) reads `(baseline − removals) ∪ additions`
+  with git-working-tree semantics — additions win on overlap, the baseline stays untouched. It is read-only,
+  and it is deliberately **not** an `IModelGroup`. The mapped reads and LINQ honour the overlay natively; caller-supplied
+  SPARQL is rewritten by `OverlayQueryRewriter` (whitelist over the dotNetRDF parse tree — property paths, `GRAPH`
+  blocks, `SERVICE`, `CONSTRUCT`/`DESCRIBE`, `FILTER EXISTS` and a caller `FROM` are **refused**), and
+  `inferenceEnabled: true` throws. Nothing silently returns unsubtracted triples. **dotNetRDF cannot serialize a
+  negated comparison faithfully** (`!(?r < 3)` → `!?r < 3`, which still parses), so `SparqlExpressionWriter` writes
+  `FILTER`/`BIND` expressions instead; `HAVING` and projected expressions come from dotNetRDF and are only checked,
+  so a negation there is refused. Every rewrite is re-parsed and compared structurally against the original.
+  Caller queries are parsed on **every** execution — see ADR-0041 for the placeholder-IRI caching optimisation
+  if that ever shows up in a profile. The SPARQL is
+  built only by `LayeredModelSparql`, whose emission rules (MINUS vs FILTER NOT EXISTS, `VALUES` before the
+  overlay, selective patterns before the wildcard) are correctness/perf requirements — see the ADR before touching them.
+  **All three graphs must be in the same store** — the overlay is one query over one dataset, so a cross-store view
+  is refused at construction (it used to fail silently in both directions); create views only via
+  `store.CreateLayeredModel`, never by constructing `LayeredModel` directly.
 - **Discovery is global static state** (0020): consumers must `MappingDiscovery.RegisterAssembly`
   / `OntologyDiscovery.AddAssembly` at startup or mapping and SPARQL prefixes silently miss.
 - **No configuration subsystem** (0011, 2.0): the `ontologies.config`/`app.config` loading,
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,7 +55,7 @@ is netstandard2.0 / net8.0 and builds cross-platform.
 
 ```bash
 dotnet build Semiodesk.Trinity.sln -c Release          # whole solution, SDK-only
-dotnet test Trinity.Tests/Trinity.Tests.csproj         # 542 passed, 7 skipped (quarantined)
+dotnet test Trinity.Tests/Trinity.Tests.csproj         # 589 passed, 7 skipped (quarantined)
 dotnet test tests/Trinity.Generator.Tests/Trinity.Generator.Tests.csproj   # 23 passed
 dotnet test tests/Trinity.Vocabulary.Tests/Trinity.Vocabulary.Tests.csproj # 29 passed
 dotnet pack Trinity/Trinity.csproj -c Release          # -> Semiodesk.Trinity.2.0.0.nupkg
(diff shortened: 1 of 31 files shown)


meberl:
## Review findings addressed — all five fixed

Thank you, this was a good catch on all counts. I reproduced every finding before touching anything, using your differential technique (empty layers vs. the plain baseline), and it found all five exactly as described. Three commits: `c40c313` fix, `395c8c6` tests, `9461050` docs.

### The three silent ones

**1. Nested UNION rewritten as a join.** Confirmed and fixed. Only `WriteChildPatterns` inspected `IsUnion`, so `WritePattern` flattened a union reached as an *alternative* of another union. Union emission now lives in one place both entry points route through. Your observation that flat `A UNION B UNION C` survived only because dotNetRDF flattens it was the key detail — it parses **left-nested**, so this was the ordinary three-way case rather than an edge one. A union carrying patterns of its own is now refused rather than quietly mishandled.

**2. Bag union duplicating solutions.** Confirmed — `COUNT(*)` returned 2 for one triple. The branches now cannot both match:

```sparql
{ { GRAPH <baseline>  { T } <guard> }
  UNION
  { GRAPH <additions> { T }
    FILTER NOT EXISTS { GRAPH <baseline> { T } FILTER NOT EXISTS { GRAPH <removals> { T } } } } }
```

The additions branch excludes whatever the baseline branch already produced, so they are disjoint by construction. The inner guard is deliberately `FILTER NOT EXISTS` and never `MINUS`: nested inside `NOT EXISTS` the pattern is tested for existence rather than joined, so a `MINUS` with no shared variables would remove nothing and invert the answer for a triple in all three graphs. You were right that an idempotent re-add is what "additions win" invites — this was the common case, not a corner.

**3. GROUP BY / ORDER BY escaping the round-trip check.** Confirmed both. Now checked like `HAVING` and the projection, following each `Child` chain, with `ORDER BY` **direction** included — reversing it changes the answer as surely as changing the expression.

### The two misfiring refusals

**4. Blank nodes.** Now refused as the unsupported form they are, naming the cause, instead of reporting a rewriter defect. You were also right that the class remark was wrong: `[ … ]` *is* a blank node, so property lists are handled only when they don't introduce one. Remark corrected.

**5. BIND relocated.** Pattern reordering is a performance requirement, but `BIND` is order-sensitive where a triple pattern is not. Reordering is now confined to contiguous runs of match patterns, so assignments keep their position.

### On the mislabelled corpus case

You identified the actual root cause, not just a cosmetic slip. `"blank node property list"` was a predicate-object list, and that is exactly why finding 4 survived. Relabelled, with a real blank-node case added — but the structural fix is `LayeredModelDifferentialTest`: with empty layers the view must answer exactly as the plain baseline does, or refuse, so **the baseline is the oracle and no expectations need authoring**. 35 shapes chosen to exercise how patterns nest rather than what they match, on every backend.

I verified it isn't vacuous by reintroducing the nested-union defect: 8 cases fail, and restoring the fix returns all 35 to green.

### Re-measured

The fix to #2 adds a nested `NOT EXISTS` to the hot macro, so the numbers had to be redone through the production path at 1M triples:

| Read | in-memory | Virtuoso | GraphDB |
|---|---|---|---|
| `GetResource` (subject-bound) | 1.31× | 0.94× | 1.29× |
| `ContainsResource` (`ASK`) | 3.45× | 1.18× | 1.04× |
| caller scan, unselective | 2.24× | 0.97× | 0.94× |
| caller selective | 4.23× | 1.47× | 2.33× |

**One honest regression:** the in-memory unselective scan went 1.50× → 2.24×. Inside the original 5× gate, and the cost of a correct answer. The ratios above 2× are on sub-millisecond absolutes where parse+rewrite dominates; mapped reads skip that path entirely. ADR-0041 is updated with these and with all five findings recorded.

### Suites

| Suite | before review | now | pre-existing failures |
|---|---|---|---|
| in-memory | 542 P | **589 P** / 7 S | 3, unchanged |
| Virtuoso | 175 P | **221 P** / 1 S | 4 inferencing, unchanged |
| GraphDB | 182 P | **228 P** / 1 S | 4 inferencing, unchanged |

Failure sets still byte-identical to `develop`.

PR #40, opened 2026-08-21
Materialize a layered view into a fourth graph

meberl:
## What

WP4. An opt-in mode: pass a materialized graph and the effective triples are kept there, so queries run **natively** against an ordinary graph.

```csharp
// rewriting mode — live, no extra storage, refuses exotic query forms
var view = store.CreateLayeredModel(baseline, additions, removals);

// materialized mode — a fourth graph holds (baseline − removals) ∪ additions
var view = store.CreateLayeredModel(baseline, additions, removals, effective);
```

The overlay macro **is** the materialization query — one `INSERT … WHERE` reusing what #37 already emits — so there's no new query machinery.

## The mode switch is two seams, not a parallel implementation

Every authored read composes its query from the dataset clause and one pattern helper. Pointing both at a single graph is all it takes for the *same* templates to run natively: the clause becomes `FROM <materialized>`, and the helper emits a bare triple pattern instead of the overlay. Caller SPARQL then skips the rewriter entirely, LINQ passes no overlay to the writer, and **inferencing becomes possible** because the store is reasoning over one ordinary graph.

That lifts every rewriting-mode refusal: unbounded property paths, the bounded paths that were merely unimplemented, `CONSTRUCT`, `DESCRIBE`, `SERVICE`, inferencing, and arbitrary caller SPARQL with no whitelist at all.

## What makes it usable rather than merely possible

Maintenance is **O(changes)**. The graph isn't recomputed after a stage — each affected triple is deleted and re-inserted *if the overlay says it belongs*. Asking the overlay is correct by construction and can't drift from the read path, where a hand-written rule per staging case could.

Measured, 1,000,000-triple baseline, in-memory:

| operation | cost |
|---|---|
| full build, with verification | 31.5 s |
| full build, unverified | 19.1 s |
| **stage a change and keep in step** | **0.7 ms** |
| native property-path query over it | 3 ms |

`Discard()` needs the full rebuild (the effective graph reverts to the baseline); a **clean** `Accept()` needs none, because the baseline becomes exactly what was materialized; a **forced** one does, because the change was applied over a baseline that had moved.

## Virtuoso has a ceiling, and doesn't say so

Measured: a single `INSERT … WHERE` succeeds at **500,000** rows and writes **zero** at **1,000,000** — reporting success either way. Its transaction log limit, silently hit.

An empty materialized graph **cannot be told apart from an empty baseline**, so a caller would read a view asserting the data does not exist. `Refresh()` therefore counts the overlay's solutions, compares them against what landed, and throws:

> Materializing this view wrote 0 triples where the overlay says there are 1,000,001. The store accepted the update and reported success, so this is a limit it did not report… on Virtuoso, `log_enable(3,1)` disables transaction logging for bulk updates.

The comparison is exact because the overlay's branches are disjoint, so each effective triple yields exactly one solution. This verification is why a build costs 31.5 s rather than 19.1 s, and it's worth the difference. Chunked materialization would lift the ceiling and is on the open list — with the note that `LIMIT`/`OFFSET` paging isn't a sound basis for it, having no stable order without `ORDER BY`.

## The two decisions ADR-0042 had deferred

- **Incremental vs dirty-marking** → incremental, by recomputing affected triples via the overlay.
- **What a stale view does on read** → the view maintains what it changes; a write straight to a layer leaves it stale and that **cannot be detected cheaply** (triple counts are unsound — swap one triple for another and the count is unchanged — and there's no change notification since ADR-0035). `Refresh()` repairs it, and the limitation is **tested rather than implied**.

## Tests

22 cases per backend. The central assertion is **equivalence**: a materialized view must answer exactly as a rewriting view over the same three layers — which makes the rewriting mode (already covered by a 56-case corpus and a differential oracle) the oracle for this one, so nothing is authored twice and the modes can't drift.

The unlocks are asserted *as* unlocks: each shows the rewriting mode throwing and the materialized mode answering, documenting the difference rather than just exercising one side. The sharpest is a transitive path over a chain whose hops straddle the layers — `a next b` in the baseline, `b next c` in the additions — so the chain exists in the effective graph and in neither layer alone.

| suite | before | now | pre-existing failures |
|---|---|---|---|
| in-memory | 602 | **624** / 7 skipped | 3, unchanged |
| Virtuoso | 234 | **256** / 1 skipped | 4 inferencing, unchanged |
| GraphDB | 241 | **263** / 1 skipped | 4 inferencing, unchanged |
| Generator / Vocabulary | 23 / 29 | 23 / 29 | none |

Exactly +22 per backend. Store suites need Docker and are excluded from the default CI job (ADR-0036); all three were run locally.

⚠️ Pre-existing failures reproducing on `develop`: in-memory `Equal` / `ResourceConstructorTest` / `EqualsTest` (a `DOTNET_ROLL_FORWARD` artefact — no net8.0 runtime installed here), and the 4 inferencing tests per store.

## One expectation of mine that was wrong

An explicit `GRAPH <effective>` block matches nothing, because the graph is selected with a plain `FROM` and is therefore the query's *default* graph rather than a named one. That's correct SPARQL, so the test asserts the forms that do work and the interface documents why that one doesn't — a caller need not know the graph's name.

## Review guide

`c3ef1b6` implementation, `6c276b9` tests, `eaeaee5` docs. Start with ADR-0042's materialization section; `LayeredModel.Refresh` / `SynchronizeMaterialized` / `VerifyMaterialized` are the parts most worth a careful read.


Trinity/Model/LayeredModel.cs
```diff
@@ -123,18 +129,35 @@ public bool IsEmpty
...
+                RequireDistinct(additions, materialized, nameof(additions), nameof(materialized));
+                RequireDistinct(removals, materialized, nameof(removals), nameof(materialized));
+            }
+
+            // The two seams that make the mode switch cheap: every authored read composes its query
+            // from this clause and from Overlay() below, so pointing both at one ordinary graph is all
+            // it takes for the same templates to run natively against the effective triples.
+            _datasetClause = materialized == null
+                ? LayeredModelSparql.NamedDatasetClause(this)
+                : "FROM " + SparqlSerializer.SerializeUri(materialized.Uri) + " ";
```
meberl, line 180: **`HasDiverged()` always returns false in materialized mode** — the stale-changeset guard from ADR-0042 is silently disabled.

Pointing `_datasetClause` at `FROM <materialized>` leaves the dataset with *no named graphs*. `HasDiverged()` (L1005-1012) composes its query from this clause but its pattern is `GRAPH <removals> { ?s ?p ?o } FILTER NOT EXISTS { GRAPH <baseline> { ?s ?p ?o } }` — with nothing in the named-graph set, neither `GRAPH` block can ever match.

Verified on the in-memory store: same three layers, same staged removal against a baseline that had moved — rewriting view returns `True`, materialized returns `False`. `Accept()` then applies the stale changeset without throwing, and without refreshing either, since only `force` triggers `Refresh`. That is exactly the merge-that-never-fails the ADR says it refuses.

The comment above is right that the two seams are the dataset clause and `Overlay()` — but `HasDiverged` is a third consumer, and it needs the layers, not the effective graph. Suggest it call `LayeredModelSparql.NamedDatasetClause(this)` explicitly rather than `_datasetClause`. Worth a grep for other `_datasetClause` readers that reason over the layers rather than the result.

Trinity/Model/LayeredModel.cs
```diff
@@ -602,11 +648,16 @@ public void UpdateResource(Resource resource, ITransaction transaction = null)
...
+            var affected = new List<string>();
+            string update = BuildStagingUpdate(resources, staged, affected);
 
             if (update != null)
             {
                 _store.ExecuteNonQuery(new SparqlUpdate(update), transaction);
+
+                // The view owns this change, so it can keep the materialized graph in step for exactly
+                // the triples it touched. That is the whole reason staging belongs to the view.
+                SynchronizeMaterialized(affected, transaction);
```
meberl, line 699: **`DeleteResource` is the missing sibling of this call** — it stages into the layers but never syncs the materialized graph, so a delete through the view is invisible to reads through the same view.

`DeleteResource(Uri, ITransaction)` (ending L1130-1137) builds its own staging update and calls `RunAtomically` directly, with no `SynchronizeMaterialized`. It already computes the affected triples as `touches`, so the fix looks like the same one line as here.

Verified: after `Materialized.DeleteResource(keep)`, `Materialized.ContainsResource(keep)` is still `True` while the rewriting view over the same layers reports `False`.

This contradicts the ADR's "the view keeps the graph in step for changes made through it" — and it fails the way this mode fails worst, since there is no signal to the caller and `Refresh()` is documented as being for out-of-band writes only.

Trinity/Model/LayeredModel.cs
```diff
@@ -521,6 +557,14 @@ public ISparqlQueryResult ExecuteQuery(ISparqlQuery query, bool inferenceEnabled
                 return ExecuteOverlayQuery(query, transaction);
             }
 
+            // Materialized: the effective graph is an ordinary graph, so a caller's query runs against
+            // it unchanged. Nothing is rewritten, so nothing has to be refused - property paths, GRAPH
+            // blocks, CONSTRUCT, DESCRIBE and inferencing all simply work.
+            if (IsMaterialized)
```
meberl, line 588: **A caller-supplied `FROM` is honoured here, which defeats the overlay.**

"Nothing is rewritten, so nothing has to be refused" holds for the constructs the rewriter refuses for want of a faithful rewrite — property paths, `GRAPH` blocks, `CONSTRUCT`/`DESCRIBE`, inferencing. A caller `FROM` is not in that category: the rewriting mode refuses it because an explicit dataset clause *overrides* the view's own, and the preprocessor here merely appends `FROM <materialized>` beside the caller's, giving a union of the two.

Verified: `SELECT ?s FROM <baseline> WHERE { ?s a <Thing> }` is refused by the rewriting view but returns a resource whose `rdf:type` is staged for removal through the materialized one — an unsubtracted triple served from a read-only subtractive view.

The same argument applies to a caller `GRAPH <baseline>` block, which is refused in rewriting mode and passes through here. Suggest keeping the dataset/named-graph refusal in both modes and lifting only the rewrite-shape refusals.

Trinity/Model/LayeredModel.cs
```diff
@@ -736,6 +790,131 @@ private static string Graph(IModel model)
...
+            RequireMaterialized();
+
+            string update = string.Join("; ",
+                string.Format("DELETE WHERE {{ GRAPH {0} {{ ?s ?p ?o }} }}", Graph(Materialized)),
+                string.Format("INSERT {{ GRAPH {0} {{ ?s ?p ?o }} }} WHERE {{ {1} }}",
+                    Graph(Materialized), LayeredModelSparql.Overlay(this, "?s", "?p", "?o")));
+
+            RunAtomically(update, transaction);
+
+            VerifyMaterialized();
```
meberl, line 858: **The check fires after the damage is committed, and only once.**

`RunAtomically` has already committed the truncated graph by the time `VerifyMaterialized()` runs, and there is no rollback and no latch. So a view that trips the Virtuoso transaction-log limit throws from `Refresh()` exactly once — and every subsequent read serves the short graph silently. `CreateLayeredModel` propagates the first throw, so the caller never gets the view; but `Discard()`/`Accept(force: true)` call `Refresh()` on a view the caller already holds, and after that one exception it keeps working and keeps lying.

The XML doc's promise — an empty graph "is reported rather than served" — only holds for the first call. Either set a poisoned flag that `RequireMaterialized()`/the read paths check, or wipe the materialized graph on failure so a later read fails loudly instead of quietly.

Trinity/Stores/StoreExtensions.cs
```diff
@@ -95,7 +113,12 @@ public static ILayeredModel CreateLayeredModel(this IStore store, IModel baselin
             RequireSameStore(store, additions, nameof(additions));
             RequireSameStore(store, removals, nameof(removals));
 
-            return store.CreateLayeredModel(baseline.Uri, additions.Uri, removals.Uri);
+            if (materialized != null)
+            {
+                RequireSameStore(store, materialized, nameof(materialized));
+            }
+
+            return store.CreateLayeredModel(baseline.Uri, additions.Uri, removals.Uri, materialized?.Uri);
```
meberl, line 134: `materialized?.Uri` silently degrades to a rewriting view when `materialized.Uri` is null — a `ModelGroup`, whose `Uri` is null by design. The overload then hands `null` to the `Uri` version, which reads it as "no fourth graph" and returns a rewriting view.

The three required layers can't hit this: `store.CreateLayeredModel(baseline.Uri, …)` throws `ArgumentNullException` on a null baseline `Uri`. So the fourth parameter is the one place where passing an unusable model gets you a *different mode* rather than an exception, and the caller who asked for materialization has no way to notice. `RequireSameStore` above already rejects the wrong store; a null-`Uri` check belongs next to it.

Trinity/Query/Sparql/SparqlQueryProvider.cs
```diff
@@ -58,11 +58,15 @@ public SparqlQueryProvider(IModel model, bool inferenceEnabled)
...
             _inferenceEnabled = inferenceEnabled;
-            _layered = model as ILayeredModel;
+            // Only a rewriting view needs the overlay woven into each pattern. A materialized one is
+            // an ordinary graph as far as the writer is concerned, selected by the dataset clause.
+            var layered = model as ILayeredModel;
 
-            if (_layered != null)
+            _layered = layered != null && !layered.IsMaterialized ? layered : null;
+
+            if (layered is LayeredModel concrete)
```
meberl: Narrowing the guard from `_layered != null` to `layered is LayeredModel concrete` also narrows *who gets refused*: a third-party `ILayeredModel` implementation used to get the `NotSupportedException` for `inferenceEnabled`, and now silently gets inferencing over a rewritten overlay — which is the case ADR-0041 says cannot be approximated.

The cause is that `RequireNoInferencing` went from static to instance to consult `IsMaterialized`, but `IsMaterialized` is on the interface. So the check can stay interface-level:

```csharp
if (layered != null && inferenceEnabled && !layered.IsMaterialized)
{
    throw ...;
}
```

or keep calling into the model but branch on `ILayeredModel`, letting `LayeredModel` be one implementation of it rather than the only recognised one.

Trinity/Model/LayeredModel.cs
```diff
@@ -167,9 +190,19 @@ private static void RequireDistinct(IModel a, IModel b, string nameA, string nam
         /// <summary>
         /// Wraps one triple pattern in the overlay.
         /// </summary>
+        /// <summary>
```
meberl, line 210: Two consecutive `<summary>` blocks — the old one-liner at L190-192 wasn't removed when this one was added. The compiler takes the first and drops the second, so the `<remarks>` explaining the mode switch is what the reader most needs and least gets.

Trinity/Query/LayeredModelSparql.cs
```diff
@@ -103,6 +103,30 @@ internal static string Overlay(ILayeredModel model, string subject, string predi
...
         }
 
+        /// <summary>
+        /// The overlay for an already-serialized triple.
+        /// </summary>
+        /// <remarks>
+        /// Used when synchronizing a materialized graph, where the triples are known ground values
+        /// rather than a pattern being built up term by term.
+        /// </remarks>
+        internal static string Overlay(ILayeredModel model, string triple)
```
meberl, line 117: This new overload makes the `<see cref="Overlay"/>` in the type's own `<remarks>` (L43) ambiguous — CS0419. Needs the signature, e.g. `<see cref="Overlay(ILayeredModel, string, string, string)"/>`.

Related doc break in the same PR: `StoreExtensions.cs:104`'s `<inheritdoc cref="CreateLayeredModel(IStore, Uri, Uri, Uri)" path="/remarks" />` no longer resolves after the signature gained the fourth parameter — CS1574 — so the `IModel` overload now inherits nothing, and the four-graph contract it was pointing at is exactly what a caller of that overload needs to read.

meberl (commented):
Code review of the materialized fourth graph. Eight findings; the first three I confirmed empirically with a throwaway probe fixture on the in-memory store (since removed).

The first two are anchored to the in-diff line that *causes* them, because the broken code itself (`HasDiverged` at L1009, `DeleteResource` at L1137) is unchanged by this PR and GitHub cannot attach a comment there.

Cleared on inspection: LINQ `Count`/`Where`/`Skip`/`Take` over a materialized view, `SerializeOffsetLimit`'s dot-less outer pattern, the `VerifyMaterialized` count queries' dataset (the preprocessor's `AddGraph` de-dupes, so the counts are right), and creating a materialized view over brand-new empty graphs.

Trinity/Model/LayeredModel.cs
meberl: Confirmed and fixed in b1b39b7.

Your diagnosis was exact, including the suggested fix. `HasDiverged` now asks for `LayeredModelSparql.NamedDatasetClause(this)` explicitly.

I took the grep you asked for. There are five readers of the field and `HasDiverged` was the only offender — the other four (`IsEmpty`, `ContainsResource`, `CreateResourceQuery`, the typed-resource read) each pair the clause with `Overlay()`, so they are result-level and correct by construction. To stop it recurring I renamed the field `_effectiveDatasetClause` and documented on it that it is only valid for a query whose patterns go through `Overlay()`; a query reasoning over the *layers* must ask for the named dataset. The name now carries the distinction that the comment about "two seams" missed.

Pinned by `HasDivergedReadsTheLayersNotTheEffectiveGraph`, plus `AcceptStillRefusesAStaleChangesetWhenMaterialized` for the consequence you identified — that `Accept()` was applying stale changesets without throwing.

Trinity/Model/LayeredModel.cs
meberl: Confirmed and fixed in b1b39b7 — it was one line short of the same call, as you said.

I audited every write path rather than just this one: `AddResource`/`CreateResource` route through `staged.Commit()` → `UpdateResources`, and all three `DeleteResources` overloads route through `DeleteResource`, so this was the single gap.

One wrinkle on the fix. `touches` is a pattern plus a `FILTER`, not a list of ground triples, and `DELETE WHERE` takes a quad pattern — a `FILTER` inside it is not legal SPARQL. So I split the filter out and added a `SynchronizeMaterialized(pattern, filter, transaction)` sibling that emits `DELETE { … } WHERE { … FILTER … }` and re-inserts what the overlay still holds. It keeps the existing principle: ask the overlay what belongs rather than derive it from the operation. For a delete the overlay provably holds nothing for that resource afterwards, but hard-coding that would be exactly the drift-prone rule the ground-triple sibling's docstring warns against.

Pinned by `DeleteThroughAMaterializedViewIsVisibleThroughIt` and `DeleteThroughAMaterializedViewFollowsTheObjectSide` — the latter for the ADR-0030 object side, where `a → b` comes from the baseline and `b → c` from the additions.

Trinity/Model/LayeredModel.cs
meberl: Confirmed and fixed in b1b39b7, and I took your framing wholesale: materialization lifts the refusals that exist *for want of a faithful rewrite*, and graph selection is not one of them.

`OverlayQueryRewriter.RequireNoGraphSelection` parses and refuses a dataset clause or a `GRAPH` block — recursing into sub-`SELECT`s and child patterns — without applying any of the other whitelist checks, so property paths, `CONSTRUCT`, `DESCRIBE` and inferencing stay lifted. The comment claiming "nothing is rewritten, so nothing has to be refused" is corrected in place rather than deleted, since it was the reasoning error.

The cost is that a materialized view now parses each caller query once per execution even though it does not rewrite it. That is the same per-execution parse ADR-0041 already documents, with the same placeholder-IRI caching available if it ever shows up in a profile — noted in the ADR rather than left implicit.

Pinned by `GraphSelectionIsRefusedInBothModes`, four cases: `FROM`, `FROM NAMED`, a `GRAPH` block, and a `GRAPH` block inside a sub-`SELECT`. Green on Virtuoso and GraphDB as well as in-memory.

I also corrected ADR-0042, which asserted the opposite — that a `GRAPH` block naming the materialized graph "matches nothing … correct SPARQL rather than a gap". That paragraph reasoned about the *materialized* graph's name and never considered a caller naming a *layer*.

Trinity/Model/LayeredModel.cs
meberl: Confirmed and fixed in b1b39b7. I took the latch rather than the wipe: it fails just as loudly, and it keeps the short graph available for diagnosis.

`_materializationFailed` is set immediately before the throw and checked in `ExecuteOverlayQuery`, which is the single funnel every read goes through — authored reads and caller SPARQL alike. `Refresh()` clears it optimistically at the start, so the rebuild is the recovery path and can still run its own verification queries; only a successful rebuild leaves it clear.

Your point that `CreateLayeredModel` propagates the first throw but `Discard()`/`Accept(force: true)` do not was the part I had not thought through — those hand back a view the caller already holds, which is precisely where the one-shot exception was worthless.

Pinned by `AShortMaterializationRefusesEveryLaterRead`. I want to flag the test's weakness honestly: it provokes the state by setting the private field via reflection, because the real trigger is a store limit somewhere between 500k and 1M rows that no test can reach in reasonable time. So it pins the latch semantics — reads refused, `Refresh()` clears — but not the detection itself, which remains covered only by the manual measurement recorded in the ADR.

Trinity/Stores/StoreExtensions.cs
meberl: Confirmed and fixed in b1b39b7 — the null-`Uri` check now sits next to `RequireSameStore`, as you suggested, and throws `ArgumentException` with `ParamName` `materialized`.

Your observation that this is the *only* parameter where an unusable model yields a different mode rather than an exception is what makes it worth a check at all: the three required layers are covered by the `Uri` overload's own null guards.

Pinned by `AMaterializedModelMustNameOneGraph`, which asserts `group.Uri` is null first so the test documents why a `ModelGroup` is the case that hits it.

Trinity/Query/Sparql/SparqlQueryProvider.cs
meberl: Fixed in b1b39b7, using your first suggestion in spirit: `RequireNoInferencing` is now a static taking `ILayeredModel`, with the instance method delegating to it, so the check is interface-level and any implementation is held to it. `IsMaterialized` is on the interface, which was the only reason the method had become an instance member.

One correction to the finding, in the interest of not claiming more than I fixed: I do not think this was reachable from outside the assembly. `SparqlQueryProvider` is `internal sealed`, and the only paths to its constructor are `Model`, `ModelGroup` and `LayeredModel`'s own `AsQueryable`. A third-party `ILayeredModel` cannot construct it, so no such implementation could actually have slipped past the concrete-type test — it was defence in depth that had quietly stopped being defence, rather than a live hole.

That is also why I have not added a test. Covering it needs a stub `ILayeredModel`, which means implementing all of `IModel`, and the test project has no mocking library. Reachable only from the test assembly via the internal `AsSparqlQueryable`, so a test would be asserting against a path no consumer has. Happy to add it if you would rather have it pinned.

Trinity/Model/LayeredModel.cs
meberl: Fixed in b1b39b7 — the stale one-liner is gone, so the `<remarks>` about the mode switch is now attached to the surviving `<summary>` and actually reaches the reader.

Trinity/Query/LayeredModelSparql.cs
meberl: Both fixed in b1b39b7, and both were real compiler warnings I should have caught before opening this — `dotnet build` was emitting CS0419 and CS1574 on the exact lines you name. The `Overlay` cref now carries the full signature, and the `inheritdoc` cref is updated to the five-parameter form.

The rest of the build is clean; the only remaining warnings in `Trinity` are the two pre-existing CS0618s for the deprecated dotNetRDF `UriLoader`/`SparqlRemoteEndpoint`.

meberl:
## Review addressed — all eight findings fixed in b1b39b7

Every finding reproduced. Three were silent wrong answers, and each is now pinned by a test that **fails without its fix**: I stashed the fixes, rebuilt the test file against the unfixed product code, and confirmed all 10 new tests go red (the 22 pre-existing materialization tests stayed green), then restored and confirmed all 32 pass. That step mattered — a test passing for the wrong reason is the specific failure this PR is about.

| # | Finding | Fix | Test |
|---|---|---|---|
| 1 | `HasDiverged()` always false when materialized | asks for the layer dataset explicitly; field renamed `_effectiveDatasetClause` | `HasDivergedReadsTheLayersNotTheEffectiveGraph`, `AcceptStillRefusesAStaleChangesetWhenMaterialized` |
| 2 | `DeleteResource` never synced the effective graph | pattern-and-filter sync sibling | `DeleteThroughAMaterializedViewIsVisibleThroughIt`, `…FollowsTheObjectSide` |
| 3 | caller `FROM`/`GRAPH` honoured when materialized | `RequireNoGraphSelection`, refused in both modes | `GraphSelectionIsRefusedInBothModes` ×4 |
| 4 | short materialization threw once, then served | latched; reads refused until a rebuild succeeds | `AShortMaterializationRefusesEveryLaterRead` |
| 5 | `materialized?.Uri` degraded to a rewriting view | null-`Uri` check beside `RequireSameStore` | `AMaterializedModelMustNameOneGraph` |
| 6 | inferencing guard narrowed to the concrete type | static, interface-typed | none — see the thread |
| 7 | doubled `<summary>` | stale one removed | — |
| 8 | CS0419 + CS1574 | crefs corrected | build is warning-clean |

**Verified:** in-memory 634 passed / 7 skipped / 3 pre-existing failures (`Equal`, `ResourceConstructorTest`, `EqualsTest` — reproduce on `develop`); **Virtuoso 163/163**; **GraphDB 163/163**. Finding 3's refusal is green on all three, which is the one I most wanted cross-store confirmation for.

### Two things I want to be explicit about rather than bury

- **Finding 6 was not reachable from outside the assembly.** `SparqlQueryProvider` is `internal sealed` and the only routes to its constructor are `Model`/`ModelGroup`/`LayeredModel`. So this was defence in depth that had stopped being defence, not a live hole — fixed, but untested, and the thread explains why a test would assert against a path no consumer has.
- **Finding 4's test provokes the state by reflection.** The real trigger is Virtuoso's transaction-log limit somewhere between 500k and 1M rows, which no test can reach in reasonable time. The test pins the latch semantics; detection itself is still covered only by the manual measurement in ADR-0042.

### Documentation corrected, not just code

ADR-0042 asserted things this review disproved, so the reasoning errors are recorded rather than quietly overwritten: the "mode switch is two seams" framing hid that a query about the *layers* is not a read of the view, and the paragraph claiming a `GRAPH` block "matches nothing … correct SPARQL rather than a gap" reasoned only about the materialized graph's own name, never about a caller naming a layer. `CLAUDE.md` claimed materialization "lifts every rewriting-mode refusal", which is now stated as lifting only those that exist for want of a faithful rewrite.

Trinity/Model/LayeredModel.cs
```diff
@@ -521,6 +576,22 @@ public ISparqlQueryResult ExecuteQuery(ISparqlQuery query, bool inferenceEnabled
...
+            // property paths, CONSTRUCT, DESCRIBE and inferencing all simply work.
+            //
+            // Graph selection is the exception, and stays refused in both modes. It is not a rewriting
+            // limitation: the view names the effective graph in its own dataset clause, and a caller
+            // FROM is appended beside it rather than replacing it, so the query would read the union of
+            // the two - including triples staged for removal. A GRAPH block reaches for the layers the
+            // view exists to combine, for the same reason.
+            if (IsMaterialized)
+            {
+                OverlayQueryRewriter.RequireNoGraphSelection(query.ToString());
```
meberl: **This refuses the view's own dataset clause, so LINQ on a materialized view is entirely broken.**

`SparqlQuery.Model`'s setter injects the view's graph into the query — `_preprocessor.AddDefaultGraph(layered.Materialized.Uri)` at `SparqlQuery.cs:89`, added by this PR. `query.ToString()` therefore already carries `FROM <materialized>` for any query whose `Model` has been assigned, and this guard reports it as the caller's.

Three ways in, all verified on the in-memory store:

**1. Every LINQ query.** `SparqlQueryProvider.cs:82` constructs `new SparqlQuery(...) { Model = _model }`, so the clause is there before execution:

```
Count()            rewriting: 1     materialized: NotSupportedException ("its own dataset clause")
Where(...).ToList() rewriting: 1     materialized: NotSupportedException
ToList()                           materialized: NotSupportedException
```

**2. Re-executing any query object.** The guard is not idempotent, because execution *mutates* the query:

```
var q = new SparqlQuery("SELECT ?s WHERE { ?s a <ex:Thing> }");
Materialized.GetBindings(q).Count();   // 3
Materialized.GetBindings(q).Count();   // NotSupportedException
q.ToString();                          // SELECT ?s FROM <...mat-effective> WHERE { ?s a <...Thing> }
```

**3. A caller who sets `.Model` themselves** — a public setter — is refused on the first execution.

Rewriting mode is unaffected (3 / 3 on re-execution), so this is a materialized-only regression, and the authored reads survive only because they call `ExecuteOverlayQuery` directly and bypass this method. `GetResources<T>()` works; `AsQueryable<T>()` does not.

The cleanest fix is for the guard to accept the view's own graph and refuse any other name — pass `Materialized.Uri` in and compare `DefaultGraphNames`/`NamedGraphNames` against it, rather than rejecting any dataset clause at all. Checking before `Model` is assigned isn't available here, since the LINQ provider assigns it at construction.

Trinity/Query/OverlayQueryRewriter.cs
```diff
@@ -93,6 +93,74 @@ internal static string Rewrite(ILayeredModel model, string queryString)
...
+        }
+
+        private static void RequireNoGraphSelection(DnrQuery.SparqlQuery query)
+        {
+            if (query.DefaultGraphNames.Any() || query.NamedGraphNames.Any())
+            {
+                throw GraphSelectionUnsupported("its own dataset clause (FROM / FROM NAMED)");
+            }
+
+            RequireNoGraphSelection(query.RootGraphPattern);
```
meberl: **`NullReferenceException` on a bare `DESCRIBE`.** `RootGraphPattern` is null for a query with no `WHERE` clause, and `RequireNoGraphSelection(GraphPattern)` dereferences it immediately at `pattern.IsGraph`.

Verified two ways — through the view, and by invoking this method directly with the string:

```
DESCRIBE <ex:keep>         materialized: NullReferenceException
DESCRIBE <ex:keep>         rewriting   : NotSupportedException (clear message)
DESCRIBE ?s WHERE { ... }  materialized: OK, 3 resources
```

So the form the PR advertises as newly working is also the form that crashes, and materialized mode gives a worse diagnostic than the rewriting mode it is supposed to improve on. A null guard here is enough — no pattern means no graph selection to find.

Worth checking `SELECT ... WHERE {}` and a bare `ASK` for the same shape while you're in here; `ASK { ?s a ex:Thing }` I confirmed is fine.

Trinity/Query/OverlayQueryRewriter.cs
```diff
@@ -93,6 +93,74 @@ internal static string Rewrite(ILayeredModel model, string queryString)
...
+            // A sub-SELECT carries its own pattern tree, and may not declare a dataset at all.
+            foreach (DnrQuery.Patterns.ITriplePattern triplePattern in pattern.TriplePatterns)
+            {
+                if (triplePattern is DnrQuery.Patterns.SubQueryPattern subQuery)
+                {
+                    RequireNoGraphSelection(subQuery.SubQuery);
+                }
+            }
+
+            foreach (DnrQuery.Patterns.GraphPattern child in pattern.ChildGraphPatterns)
```
meberl, line 212: **The walk misses `GRAPH` blocks nested inside a `FILTER EXISTS` / `NOT EXISTS`.** dotNetRDF keeps those patterns in the filter's expression tree, not in `ChildGraphPatterns`, so neither this loop nor the `SubQueryPattern` loop above reaches them.

Verified — the same query, refused by one mode and silently answered by the other:

```
SELECT ?s WHERE { ?s a ex:Thing . FILTER EXISTS { GRAPH <baseline> { ?s ex:label 'gone' } } }
  rewriting   : NotSupportedException ("a filter that matches a graph pattern of its own")
  materialized: [] 

SELECT ?s WHERE { ?s a ex:Thing . FILTER NOT EXISTS { GRAPH <additions> { ?s ?p ?o } } }
  materialized: keep, both, added   <- 'both' and 'added' *are* in the additions graph
```

The good news is that it does not leak removed triples: the dataset clause is a bare `FROM <materialized>`, so the named-graph set is empty and the nested `GRAPH` block matches nothing. But that makes it a silent wrong answer rather than a refusal — the caller asks about a named graph and is told it is empty. `GRAPH ?g` inside a filter goes the same way, while at the top level it is correctly caught.

This also breaks the fixture's stated central assertion, that the two modes answer alike. Walking `pattern.Filter`'s expression tree for `ExistsFunction` would close it; `OverlayQueryRewriter` already has the machinery, since rewriting mode detects exactly this case.

Trinity/Model/LayeredModel.cs
```diff
@@ -736,6 +827,182 @@ private static string Graph(IModel model)
...
+        {
+            if (!IsMaterialized)
+            {
+                return;
+            }
+
+            string update = string.Join("; ",
+                string.Format("DELETE {{ GRAPH {0} {{ {1} }} }} WHERE {{ GRAPH {0} {{ {1} . {2} }} }}",
+                    Graph(Materialized), pattern, filter),
+                string.Format("INSERT {{ GRAPH {0} {{ {1} }} }} WHERE {{ {2} {3} }}",
```
meberl: This calls the ground-triple `Overlay` overload with `"?s ?p ?o ."` — a fully-variable pattern — which is the one input that overload's contract excludes ("known ground values rather than a pattern"). Two consequences, both performance rather than correctness:

1. The baseline branch gets `FILTER NOT EXISTS` unconditionally, via `GroundGuard`. That is always *correct*, but `LayeredModelSparql.Guard` documents `MINUS` as the 18x-faster choice precisely when the pattern binds a variable — which it does here. The overload has no way to pick, since it takes an opaque string.
2. `filter` sits *outside* the overlay's `{ … UNION … }` group, so nothing guarantees the engine pushes the subject/object restriction into the branches. A delete could scan the whole baseline instead of the handful of triples touching the resource, which would undercut the O(changes) claim this mode rests on.

Both are worth a measurement at the 1M baseline you already have set up rather than a guess. If either bites, the fix is a variable-aware overload — or interpolating the filter into each branch alongside the guard, the way the ground-triple sibling gets its selectivity from the triple itself.

meberl (commented):
Re-reviewed at `b1b39b7`. **All eight findings are genuinely closed**, and I verified the three substantive ones empirically rather than taking the commit message for it (throwaway probe fixture on the in-memory store, since removed — tree is clean):

| # | Verified |
|---|---|
| `HasDiverged()` | rewriting `True` / materialized `True` — was `True`/`False` |
| `DeleteResource` | `Materialized.ContainsResource` now `False`, and the effective graph no longer holds the resource's triples |
| caller `FROM` | refused in both modes |
| latch | reads refused while latched (`ContainsResource`, `IsEmpty`, `GetResource`); `Refresh()` clears it |
| null `Uri` | refused |
| inference guard | interface-level again |
| doc comments | CS0419 / CS1574 both gone, duplicate `<summary>` gone |

The recursion also correctly catches `GRAPH` under `MINUS`, `OPTIONAL` and sub-`SELECT`, and `CONSTRUCT` / `DESCRIBE ?s WHERE` / unbounded property paths all work in materialized mode as advertised.

**But the graph-selection guard introduced three new problems, one of them a broad regression.** The first is the headline: the guard cannot tell the view's *own* injected dataset clause from a caller's, so LINQ is now entirely broken on a materialized view. It slipped through because the materialization suite has no `AsQueryable` test — the rewriting suite has three (`LayeredModelTest.cs:415,423,434`) and the stated "equivalence is the central assertion" doesn't yet extend to the LINQ path.

Suggest adding the equivalence assertion for `AsQueryable<T>()` and for a re-executed query object; both are one-liners in the existing fixture and both fail today.

Trinity/Query/LayeredModelSparql.cs
```diff
@@ -103,6 +103,30 @@ internal static string Overlay(ILayeredModel model, string subject, string predi
             return "{ " + baselineBranch + " UNION " + additionsBranch + " }";
         }
 
+        /// <summary>
+        /// The overlay for an already-serialized triple.
+        /// </summary>
+        /// <remarks>
+        /// Used when synchronizing a materialized graph, where the triples are known ground values
```
meberl: This is now inaccurate: the overload's only other caller passes `"?s ?p ?o ."` from the new pattern-based `SynchronizeMaterialized`, so "known ground values rather than a pattern being built up term by term" is exactly what it no longer always receives. Given the file's careful reasoning about ground-vs-variable choosing `MINUS` over `FILTER NOT EXISTS`, this comment is the one a future reader would rely on to avoid the mistake.

Same class of thing in `Parse`'s catch block: its message says the query "cannot be rewritten to honour the layered overlay", which no longer fits the materialized path where nothing is rewritten and only the graph-selection check runs. Also worth noting in the ADR that materialized mode now requires the strict parser to accept the query, so Trinity's wider extended syntax is refused there too — a restriction this mode did not have before this commit.

Trinity/Model/LayeredModel.cs
meberl: Confirmed and fixed in 7d89d54, using your suggestion: the guard takes the effective graph's URI and refuses only *other* names. The check is "no graph but this one", not "no graph at all" — and I had not seen that the view's own clause is present by the time `ToString()` runs, let alone that execution mutates the query object.

Your diagnosis of why it slipped through is the part I want to acknowledge properly. The fixture's stated central assertion is mode equivalence, and it did not extend to the LINQ path at all — so the one guard I added could break every LINQ query on a materialized view and still show a green suite. That is a fixture design failure, not bad luck.

Added, as you suggested, plus one more:

- `LinqCountAgreesBetweenTheModes`, `LinqWhereAgreesBetweenTheModes`, `LinqEnumerationAgreesBetweenTheModes` — the setup now seeds mapped `Agent`s across the layers (one baseline, one staged addition, one whose name is staged for removal) so LINQ has something to select
- `AQueryObjectCanBeExecutedTwice` — the idempotence case
- `ACallerMayAssignTheModelItself` — the public setter, your third route in

All four fail on b1b39b7.

Trinity/Query/OverlayQueryRewriter.cs
meberl: Confirmed and fixed in 7d89d54 with a null guard, as you suggested.

I took your advice to check the neighbouring shapes, and they are *not* the same: `SELECT ?s WHERE {}` and a bare `ASK {}` get an empty but **non-null** `RootGraphPattern` from dotNetRDF, so only a query with no `WHERE` clause at all has the null. Both are in `AQueryWithoutAPatternDoesNotCrash` anyway, since the distinction is not obvious enough to leave unpinned.

Also added `ABareDescribeReturnsTheEffectiveTriples`, which asserts `DESCRIBE <gone>` does not report the `rdf:type` staged for removal — the crash was hiding the fact that nothing checked the answer was actually subtracted.

Trinity/Query/OverlayQueryRewriter.cs
meberl: Confirmed and fixed in 7d89d54. You were right that the machinery was already there — `RequireNoGraphAccess` walks exactly this tree in rewriting mode; I simply did not reuse it.

One deliberate difference. Rewriting mode refuses `EXISTS` *outright*, because it cannot weave the overlay into it. A materialized view can evaluate one perfectly well — it is an ordinary graph — so refusing it would be over-refusal of the kind this whole guard is supposed to avoid. The new walk recurses into `ExistsFunction.Pattern` / `GraphPatternTerm.Pattern` and refuses only the graph selection inside. `AFilterExistsWithoutAGraphBlockIsAllowedWhenMaterialized` pins both halves: allowed materialized, still refused rewriting.

Coverage is `FILTER EXISTS`, `FILTER NOT EXISTS`, `GRAPH ?g` in a filter, `!EXISTS` nested one level deeper, and `BIND(EXISTS {…} AS ?x)` — plus filters reached via `UnplacedFilters` and inline `FilterPattern`/`IAssignmentPattern`, since dotNetRDF holds them in more than one place.

Trinity/Model/LayeredModel.cs
meberl: Measured, as you asked — and the measurement says your point 2 was right and my first fix was wrong.

I initially did the obvious thing: interpolate the filter into each branch, and use the variable-aware `Guard` so the baseline branch gets `MINUS`. On a 1,000,000-triple in-memory baseline:

| shape of the re-insert | cost |
|---|---|
| ground guard, filter outside the `UNION` (as reviewed) | 3,891 ms |
| `MINUS`, filter interpolated into both branches | 3,695 ms |
| **two bound patterns, no filter** | **0–3 ms** |

So moving the filter inward buys 1.05x on something that needed 1000x. The reason is more basic than filter placement: `?s ?p ?o` with `FILTER (?s = <r> || ?o = <r>)` **cannot be answered from an index at all**, so the engine enumerates the graph and tests every row. The subject side and the object side have to be separate *bound* patterns.

`DeleteResource` end to end, same baseline: **12,528 ms → 3 ms**. So the O(changes) claim this mode rests on was genuinely broken, not merely at risk.

I reverted the `constraint` parameter I had added to `Overlay` — keeping it would advertise the shape that does not work — and rebuilt both the sync and, since it had the identical shape, the **staging** update in `DeleteResource`. That part predates this PR (it came in with #38), but the same measurement condemns it and it is what the 12.5 s was mostly made of.

Your point 1 is fixed as a side effect: with a bound subject or object the term-wise overload sees the variables and picks `MINUS`.

Trinity/Query/LayeredModelSparql.cs
meberl: Both corrected in 7d89d54.

The `Overlay(model, triple)` doc now states ground terms as a **precondition** rather than a description of usage, and says what goes wrong otherwise — `FILTER NOT EXISTS` unconditionally, forgoing the `MINUS` anti-join — with a pointer to the term-wise overload. As it happens the inaccuracy is gone too: the pattern-based caller you spotted is now two bound patterns going through the term-wise overload, so this one really does only receive ground triples again. You were right that this comment is the one a future reader would rely on.

`Parse`'s message no longer claims a rewrite; it says the query cannot be *checked* against the overlay, which is either rewriting it or, on a materialized view, confirming it selects no graph of its own.

The ADR note is added, and I am glad you asked for it — it is the one restriction this mode *adds*: materialized mode now requires the strict SPARQL parser to accept the query, so Trinity's wider extended syntax is refused there, which it was not before this commit. Recorded as a real narrowing traded for not serving unsubtracted triples, rather than buried.

meberl:
## Round 2 addressed in 7d89d54 — including a performance claim that was simply false

All five closed. Same verification discipline: stash the fixes, rebuild the tests against the unfixed code, confirm they go red. **13 of 16 new tests fail on b1b39b7.**

The three that pass pre-fix, accounted for rather than glossed:
- `AQueryWithoutAPatternDoesNotCrash("empty WHERE")` and `("bare ASK")` — you asked me to check these for the same shape. They are **not** the same: dotNetRDF gives them an empty but non-null `RootGraphPattern`, so only a query with no `WHERE` clause at all has the null. Kept as tests because the distinction isn't obvious.
- `AFilterExistsWithoutAGraphBlockIsAllowedWhenMaterialized` — an over-refusal guard, so passing before is correct. It proves the new filter walk didn't start refusing plain `EXISTS`.

### The headline regression, and why it reached you

Fixed as you suggested: the guard takes the effective graph's URI and refuses only *other* names — "no graph but this one". I had not seen that assigning `Model` injects the clause, nor that execution mutates the query object.

Your point about *why* it slipped through is the more useful half. The fixture's stated central assertion is mode equivalence, and it didn't extend to the LINQ path at all — so a guard that broke every LINQ query on a materialized view still showed green. That's a fixture design failure. It now asserts equivalence for `Count`, `Where` and enumeration, with mapped `Agent`s seeded across the layers, plus your re-execution and caller-assigns-`Model` cases.

### Finding 4 — the measurement contradicted my first fix

I did the obvious thing first: interpolate the filter into each branch and use the variable-aware `MINUS`. At 1M triples:

| shape of the re-insert | cost |
|---|---|
| ground guard, filter outside the `UNION` (as reviewed) | 3,891 ms |
| `MINUS`, filter interpolated into both branches | 3,695 ms |
| **two bound patterns, no filter** | **0–3 ms** |

Filter placement buys **1.05x on something that needed 1000x**. The cause is more basic: `?s ?p ?o` with `FILTER (?s = <r> || ?o = <r>)` can't use an index at all. `DeleteResource` end to end went **12,528 ms → 3 ms**, so the O(changes) claim was broken, not merely at risk. I reverted the `constraint` parameter I'd added — keeping it would advertise the shape that doesn't work — and applied the same fix to the **staging** update, which had the identical shape and accounted for most of the 12.5 s. That part predates this PR (#38), but the same measurement condemns it.

### The rest

- **Bare `DESCRIBE` NRE** — null guard, plus `ABareDescribeReturnsTheEffectiveTriples`, since the crash was hiding that nothing asserted the answer was actually subtracted.
- **`GRAPH` inside `FILTER EXISTS`** — the walk now recurses into `ExistsFunction.Pattern`/`GraphPatternTerm.Pattern`, and through `UnplacedFilters` and inline filter/assignment patterns too. One deliberate asymmetry: rewriting mode refuses `EXISTS` outright, a materialized view evaluates it happily, so only the graph selection inside is refused — pinned from both sides.
- **Doc accuracy** — the ground-triple contract is a precondition again (and true again), `Parse`'s message no longer claims a rewrite, and the ADR records the restriction this mode *adds*: materialized mode now needs the strict parser, so Trinity's wider extended syntax is refused there.

**Verified:** in-memory 650 passed / 7 skipped / 3 pre-existing failures; **Virtuoso 179/179**; **GraphDB 179/179**.

Worth noting the shape of these two rounds: round 1 was defects in the feature, round 2 was defects in round 1's fix. The guard I added to close a silent wrong answer introduced a broader regression than the bug it fixed. Both times the gap was the same — a claim in the ADR that no test held me to.

meberl:
## Forward-merged `develop` (`92f5eb1`)

`develop` had moved 24 commits ahead — ADR-0043 (Fuseki revival), ADR-0044 (store suites in CI), ADR-0045 (in-memory RDFS inferencing). Branch is now current and mergeable.

**Conflicts:** two, both in `CLAUDE.md`, both docs where develop's text strictly supersedes mine — the Fuseki "4/86 upstream connector bug" turned out to be a misdiagnosis (the container never created a dataset), and the store suites now run in CI rather than being excluded. Took develop's side wholesale. `SparqlQuery.cs` and `doc/adr/README.md` auto-merged; our edits were in different regions.

**One thing I checked before merging rather than after:** two develop commits are titled *"walk every expression a query carries, not only its filters"* and *"refuse named-graph access wherever it hides, by whitelist"* — which is, almost word for word, the pair of defects your round-2 review found here. They're in `RdfsEntailment.cs`, so no conflict: the same two mistakes were made and found independently in the inferencing work. Worth knowing that this failure mode has now appeared twice in two unrelated places.

### Two cells the merge created, now covered (`f02af0c`)

- **Fuseki had every layered fixture except materialization**, because it gained the other four before materialization existed. Filled — green first run, 315 tests.
- **ADR-0042 claims materialization lifts the inferencing refusal**, and until ADR-0045 that was unverifiable on the in-memory store: the flag was accepted and then quietly did nothing. So the claim is now split by what is actually store-independent — the shared fixture asserts a materialized view *accepts* an inference-enabled read where a rewriting one refuses it (Trinity's refusal, so every backend), and the in-memory subclass asserts the substantive part: entailments come from the *effective* triples, so a staged removal withdraws what it entailed. That test carries a control showing the same query over the baseline *does* entail it, so it cannot pass for the wrong reason.

### Verified on the merged tree

| suite | result |
|---|---|
| in-memory | 713 passed, 3 skipped |
| Virtuoso | 301 / 302 |
| GraphDB | 312 / 313 |
| Fuseki | 314 / 315 |
| generator | 23 / 23 |
| vocabulary | 29 / 29 |

The 3 in-memory failures (`Equal`, `ResourceConstructorTest`, `EqualsTest`) are the .NET 10 roll-forward `UriRef` defects that **PR #45** addresses. I confirmed them on a clean `origin/develop` worktree — identical three, so the merge introduced nothing. `CLAUDE.md`'s "0 failed on net8.0" is precise; we run under `DOTNET_ROLL_FORWARD=LatestMajor` because no net8.0 runtime is installed here, which is exactly what surfaces them.

Also refreshed the store-suite counts in `CLAUDE.md` (this branch's tests invalidated develop's figures) and the in-memory skip count, which ADR-0044/0045 took from 7 to 3.

meberl:
## Forward-merged `develop` again (`e5c745c`)

PR #45 (`UriRef` on .NET 10) had landed — 6 commits. **Clean merge, no conflicts**: develop's changes to `SparqlQueryProvider.cs` (a `Uri`→`UriRef` bridge in the binding converter, line ~284) and mine (the inferencing guard, line ~58) are in different regions, and the two `CLAUDE.md` additions are in different sections. Verified both survived: develop's `TRIN007` row and this branch's materialization mental-model entry are both present.

### The three failures I'd been reporting all session are gone

| suite | before this merge | after |
|---|---|---|
| in-memory | 710 passed, **3 failed** | **733 passed, 0 failed**, 3 skipped |
| Virtuoso | 301 / 302 | 303 / 304 |
| GraphDB | 312 / 313 | 314 / 315 |
| Fuseki | 314 / 315 | 316 / 317 |
| generator | 23 / 23 | 26 / 26 |
| vocabulary | 29 / 29 | 29 / 29 |

`Equal`, `ResourceConstructorTest` and `EqualsTest` were the .NET 10 `IEquatable<Uri>` defect, and #45 fixed it. So this branch is now **fully green on every suite** under `DOTNET_ROLL_FORWARD=LatestMajor`, with no caveat to report. Store-suite counts in `CLAUDE.md` refreshed to the measured figures (`8b4fea6`).

### One thing I noticed but deliberately did not touch

`doc/known-test-failures.md` now says the suite "is run under `DOTNET_ROLL_FORWARD=LatestPatch|Major|LatestMajor` — .NET 8, 9 and 10 — because passing on one runtime is what let this hide."

`.github/workflows/ci.yml` installs `8.0.x` and `10.0.x` but never sets `DOTNET_ROLL_FORWARD`, so CI still executes the `net8.0` test projects on the 8.0 runtime only. The multi-runtime run described there is a manual step, not something CI enforces — which is the same shape of problem as the ADR claims this PR spent two review rounds on: the sentence that would have caught the regression describes a check nothing performs.

I left it alone because it is develop's text and unrelated to materialization; folding it in here would mix concerns. Worth its own small change — a matrix over `DOTNET_ROLL_FORWARD` in the `build` job would make the claim true, and it is nearly free since both runtimes are already installed.

PR #42, opened 2026-08-21
Revive the Fuseki store: the 4/86 was a missing dataset, not a connector bug

meberl:
## The 4/86 was never a connector bug

`Trinity.Fuseki` has been recorded as **4/86, blocked on an upstream `FusekiConnector` bug** ("POSTs
`/ds/query`, which the server 404s") in ADR-0036, ADR-0009, ADR-0041, `README.md` and `CLAUDE.md`.
`FusekiStore` even carried a `TODO` agreeing with it.

That diagnosis was wrong. Probing a live container:

```
$ curl -u admin:test http://localhost:PORT/$/datasets
{ "datasets" : [ ] }
```

`FusekiContainer` created its dataset with `FUSEKI_DATASET_1=ds`. That variable belongs to
`secoresearch/fuseki`; `stain/jena-fuseki` ignores it silently. **No dataset was ever created**, so
every path under `/ds/*` returned 404 — GET and POST alike, query, update and Graph Store Protocol.
The connector was never at fault, and neither was the query URL.

| request | before | after provisioning |
|---|---|---|
| `POST /ds/query` | 404 | 200 |
| `GET /ds/query` | 404 | 200 |
| `POST /ds/update` | 404 | 200 |
| `PUT /ds/data?graph=…` | 404 | 201 |
| `DELETE /ds/data?graph=…` | 405 | 204 |

It cost us a backend: Fuseki was excluded from the ADR-0039/0041/0042 test work on the strength of the
claim, leaving it the only store with no coverage for write semantics, numeric round-trips or layered
views — and the only one still running a hand-written copy of the store suite.

## Results

| suite | before | after |
|---|---|---|
| **Fuseki** | 4 passed / 82 failed | **248 passed / 0 failed / 1 skipped** |
| GraphDB | 241 / 4 / 1 | 241 / 4 / 1 (unchanged) |
| Virtuoso | 234 / 4 / 1 | 234 / 4 / 1 (unchanged) |
| in-memory | 602 / 3 / 7 | 604 / 3 / 7 (+2 regression tests) |
| generator | 23 | 23 |
| vocabulary | 29 | 29 |

GraphDB and Virtuoso were re-run because this PR edits `StoreBase` and deletes per-store overrides in
favour of it. Both are identical to their documented failure sets **in identity, not merely in count**.
The three in-memory failures are the pre-existing `UriRef` cases that fail only on a rolled-forward
.NET 10 runtime (`doc/known-test-failures.md`).

## Reviewers: two changes reach beyond Fuseki

**`Trinity/Query/SparqlPreprocessor.cs` — the duplicated `FROM NAMED` clause.** Every layered read built
its query with the overlay's three `FROM NAMED` graphs already in the text, then set `query.Model`, whose
setter adds the same graphs through the preprocessor. The preprocessor skips a graph it has already
recorded — but it never recorded a `FROM NAMED` one. The tokeniser splits `FROM NAMED <g>` into three
tokens (`FROM`, `NAMED`, URI), and the check looked only for `Token.FROM` (115) and `Token.FROMNAMED`
(117); `FROMNAMED` never appears in that position. Plain `FROM` matched and worked, so only the layered
path — the sole emitter of `FROM NAMED` — was affected.

Virtuoso and GraphDB tolerate the repeated clause, which is why their layered suites passed throughout
ADR-0041 and 0042. Jena rejects it (`URI already in named graph set`, HTTP 400). **105 of Fuseki's 106
initial failures were this one bug.** Two regression tests, both verified to fail when the defect is
reintroduced. The pre-existing `DatasetClauseNamesAllThreeGraphsAndNeverMergesThem` passed the whole
time because it asserted the clause in isolation rather than the query after the model is assigned.

**`Trinity/Stores/StoreBase.cs` — a null guard on `ContainsModel(IModel)`**, three lines. GraphDB had
already written this guard into its own override; deleting Fuseki's override otherwise inherited a
`NullReferenceException`. Caught by a new test, not by reading.

## Adapter fixes

Wrong answers: `ContainsModel(Uri)` called `HasGraph`, **discarded the result** and returned `false`
unconditionally · `CreateModelGroup(params IModel[])` always returned an **empty group** (the
ADR-0009/0019 bug — declared `new` rather than `override`, which is why it never bit; deleting it is the
whole fix) · `IsReady` was hardcoded `true` and stayed true after `Dispose` · `TryParse` had no `Trig`
case so TriG was parsed as **RDF/XML** · `Read(Uri, Uri, …)` silently rejected `https` · a password-only
config dropped its credentials · the provider defaulted `dataset` to the literal `"dataset"` and never
validated it, so omitting `dataset=` produced a store that connected, reported ready and 404'd on
everything — *the same failure mode as the container bug above*.

Dead code: `FusekiSparqlQueryResult` (zero references) · `Properties/AssemblyInfo.cs` (its
`[assembly: Guid]` was byte-identical to `Trinity.Virtuoso`'s — two assemblies claiming one typelib
GUID) · the `_parser` field and the commented-out line that was its only consumer · four overrides that
only repeated `StoreBase`. `ListModels` moved off the obsolete `ListGraphs()`.

**`UpdateResource` is deliberately unchanged.** It differs from `StoreBase` only in emitting
`INSERT DATA` for a new resource instead of allocating a blank-node id. The three blank-id tests the
generic suite newly runs against Fuseki pass as it stands, so there is no measured reason to change the
SPARQL it emits.

## Two things this requires

- **Fuseki 5.x.** Jena 4.0.0 answers HTTP 500 *"Not a valid UUID string"* to any query naming a
  `urn:uuid:` IRI — verified against a live server, and not specific to `VALUES`: a plain triple pattern
  and a `FILTER` equality fail identically while `INSERT` succeeds. Since `Model.CreateResource()` mints
  `urn:uuid:` by default, such a resource was writable but permanently unreadable. Fixed upstream; the
  test image is pinned to `stain/jena-fuseki:5.1.0`. **This is a real constraint on consumers, not just
  on tests.**
- **Inferencing is not supported.** A Jena reasoner is a property of the dataset, so it applies to every
  query or none; there is no per-query switch. `ExecuteQuery` ignores `IsInferenceEnabled` — permitted by
  ADR-0022, and now documented on the method instead of silent. The four shared inferencing tests are
  `Assert.Inconclusive`, following the `Trinity.Tests/Store/DotNetRDF/` precedent. Giving the dataset a
  reasoner would make them pass and make `inferenceEnabled: false` quietly lie.

## Docs

ADR-0043 records the investigation, the capability decisions and the core defect. Six documents asserted
the misdiagnosis; it is corrected where stated as fact and left visible as history where a passage
reports what was measured at the time. ADR-0022 and ADR-0028 still said Fuseki/GraphDB
`BeginTransaction` returns `null` — ADR-0039 changed that to `NoOpTransaction`. `known-test-failures.md`
had no store section at all and now tabulates what each backend skips and why.

## Not done here, recorded as follow-ups in ADR-0043

Repo-wide and present identically in `GraphDBStore` and `dotNetRDFStore`, so they do not belong in a
Fuseki change: the `leaveOpen` parameter is ignored because the `using (TextReader …)` disposes the
stream regardless · an unreachable relative-`Uri` branch · the TriG loop that deletes the *target* graph
once per iteration · `Write(…)` silently writing nothing when the graph is absent · three verbatim copies
of `TryParse`. Also worth promoting `ContainsModelTest` to a shared fixture, and looking at GraphDB's
four inferencing failures — its container *does* configure a `rdfsplus-optimized` ruleset, so those are
an unresolved issue there rather than a declared limitation.

## Verifying locally

Needs a Docker daemon. These suites stay out of the default CI job, unchanged from ADR-0036.

```bash
dotnet build Semiodesk.Trinity.sln -c Release
dotnet test tests/Trinity.Tests.Fuseki/Trinity.Tests.Fuseki.csproj -c Release
dotnet test tests/Trinity.Tests.GraphDB/Trinity.Tests.GraphDB.csproj -c Release
dotnet test tests/Trinity.Tests.Virtuoso/Trinity.Tests.Virtuoso.csproj -c Release
```

To confirm the readiness probe does its job, comment out `CreateDatasetAsync` in `FusekiContainer` —
the fixture must fail `OneTimeSetUp` naming the cause, not produce 80-odd 404s mid-run.


Trinity/Query/SparqlPreprocessor.cs
```diff
@@ -297,6 +291,26 @@ private IToken ProcessUri(IToken token)
...
+        /// (FROM, NAMED, URI), so the token preceding the URI of a <c>FROM NAMED</c> clause is NAMED
+        /// and never FROMNAMED. Without this, a <c>FROM NAMED</c> graph was not recorded in
+        /// <see cref="DefaultGraphs"/>, and <see cref="AddGraph"/> - which skips a graph already in
+        /// that set - would append a second, identical dataset clause. Plain <c>FROM</c> was
+        /// unaffected, which is why only the layered read path (the one emitter of
+        /// <c>FROM NAMED</c>) ever produced a duplicate. Virtuoso and GraphDB tolerate the
+        /// repetition; Jena rejects it outright with "URI already in named graph set".
+        /// <c>NAMED</c> appears nowhere else in the SPARQL grammar, so accepting it is unambiguous.
+        /// </remarks>
+        private bool FollowsDatasetKeyword =>
```
meberl, line 323: **Regression: `FROM NAMED` graphs are recorded into the default-graph set.**

Accepting `Token.NAMED` here is the right diagnosis of the duplicate dataset clause, but the graph then lands in the same `DefaultGraphs` set that `AddGraph` de-dupes against. So for a query whose text already names a graph via `FROM NAMED`, `AddDefaultGraph` is now silently skipped for that graph.

Confirmed against the built `Semiodesk.Trinity.dll` + dotNetRDF 3.5.2:

```csharp
var q = new SparqlQuery("SELECT * FROM NAMED <http://ex/g> WHERE { ?s ?p ?o }");
q.Model = model;   // the model for http://ex/g
// emitted SPARQL now carries no FROM clause at all
```

Empty default graph, zero rows — where before this change it emitted both clauses. The same conflation makes `Bind` on a `FROM @model` parameter throw "already set".

The fix wants a separate named-graph set: keep `DefaultGraphs` for `FROM` operands, add e.g. `NamedGraphs` for operands following `FROM NAMED`, and have `AddGraph`/`AddDefaultGraph` de-dupe against the matching one. Worth a regression test for the plain-`FROM NAMED` + `Model` combination — the layered path is the only current emitter, which is exactly why nothing caught this.

Trinity/Stores/StoreBase.cs
```diff
@@ -87,7 +87,9 @@ public virtual void RemoveModel(IModel model)
         [Obsolete("This method does not list empty models. At the moment you should just call GetModel() and test for IsEmpty")]
         public virtual bool ContainsModel(IModel model)
         {
-            return ContainsModel(model.Uri);
+            // A null model contains nothing; it is not an error. GraphDB already guarded this in its
+            // own override, so the guard belongs here rather than being repeated per store.
+            return model != null && ContainsModel(model.Uri);
```
meberl, line 93: **The centralized guard doesn't reach Virtuoso.**

The comment says the guard "belongs here rather than being repeated per store", but `Trinity.Virtuoso/VirtuosoStore.cs:206` still overrides `ContainsModel(IModel)` with an unguarded `ContainsModel(model.Uri)`. So Virtuoso keeps throwing `NullReferenceException` on a null model while every other store now returns `false` — the inconsistency the centralization was meant to remove.

The new test that pins this behavior lives in the Fuseki-only fixture, so nothing exercises the Virtuoso path. Either drop Virtuoso's `IModel` override (its `Uri` overload is what actually differs) or add the same guard there, and move the test into a shared fixture.

Also: ADR-0043 states Virtuoso implements `ContainsModel` correctly — that's true for the `Uri` overload but wrong for the `IModel` one, so the ADR line needs a qualifier.

meberl (commented):
Code review of the Fuseki revival. Built the solution (0 errors) and ran the in-memory suite: 604 passed / 3 failed / 7 skipped — the 3 failures are the pre-existing `UriRef`/`ResourceConstructor` .NET 10 roll-forward artifacts the ADR already documents, unrelated to this diff.

The container diagnosis and the fix it unblocked both look right, and the twelve shared fixtures are a real coverage win. Three findings below; the first is the one I'd block on, since it is a core-path regression outside Fuseki.

Checked and cleared: the `Token.NAMED` acceptance is unambiguous in practice (`USING NAMED` is the only other grammar position and Trinity emits none); `OverlayQueryRewriter` refuses any caller dataset clause, so the layered path cannot collide with a caller's; `Model.ExecuteQuery`'s `query.Model == null ||` short-circuit leaves that guard unaffected; `ContainsModel(Uri)` → `HasGraph` is a HEAD request that correctly returns false on 404; `FusekiConnector` really does derive `/query` from `/data`, so the deleted TODO was correct to delete; the getter-only `IsReady` override is legal C# and nothing assigns it here (note `GraphDBStore.cs:113` has the same shape *and* assigns `IsReady = false`, a pre-existing silent no-op outside this diff); `config["uid"]` → `config[userKey]` is cosmetic since `userKey == "uid"`. ADR-0043 already lists the remaining adapter issues I spotted (`leaveOpen` ignored, the TriG loop deleting its target graph per iteration, `Write` silently writing nothing) as deliberate follow-ups.

Trinity.Fuseki/FusekiStore.cs
```diff
@@ -248,9 +273,11 @@ public override IModel GetModel(Uri uri)
         /// <returns>All handles to existing models.</returns>
         public override IEnumerable<IModel> ListModels()
         {
-            foreach (var graph in Connector.ListGraphs())
+            // ListGraphNames rather than the obsolete ListGraphs; it yields the names as strings
+            // (ADR-0038).
+            foreach (var graph in Connector.ListGraphNames())
```
meberl, line 278: **`ListModels()` throws on a blank-node-named graph.**

`ListGraphNames()` yields blank-node graph names as bare ids, not IRIs, and `new UriRef("b0")` throws `UriFormatException` — so the enumeration dies mid-stream instead of skipping the graph.

Confirmed on a dataset holding one IRI-named and one bnode-named graph: `ListGraphNames()` returns `http://ex/g1`, `b0`, while the obsolete `ListGraphs()` returns only `http://ex/g1`. The switch to `ListGraphNames` (correct per ADR-0038) is what exposes this.

Since `IModel` is IRI-keyed, skipping is the right behaviour:

```csharp
foreach (var graph in Connector.ListGraphNames())
{
    if (!Uri.IsWellFormedUriString(graph, UriKind.Absolute))
    {
        continue;   // blank-node graph name; not addressable as an IModel
    }

    yield return new Model(this, new UriRef(graph));
}
```

Any store whose `ListModels` goes through `ListGraphNames` has the same exposure, so it may belong in a shared helper.

Trinity/Query/SparqlPreprocessor.cs
meberl: Confirmed and fixed in 01d3b6b — thank you, this was a regression I introduced and it was worse than the duplicate it was fixing.

Reproduced your case exactly, then measured the consequence: the same query returns **1 row** without `FROM NAMED <g>` and **0 rows** with it. Silent, where the duplicate at least made Jena shout.

One more consequence worth recording: `GraphDBStore.cs:269` feeds `GetDefaultModels()` back in as `FROM` clauses when adding its inference graph, so a caller's `FROM NAMED` graph would have been merged into the default graph on that path too.

Fixed the way you suggested — a separate `NamedGraphs` set, with `AddGraph` de-duping against the set matching the clause it is adding. `GetDefaultModels()` deliberately keeps reporting `FROM` operands only, since both callers read it as default-graph membership; its summary claimed it returned `FROM NAMED` too, which it never did, so that is corrected. My earlier `DeclaredGraphsAreRecordedForFromNamedAsWellAsFrom` asserted the wrong contract and is replaced by `DeclaredDefaultGraphsExcludeFromNamedOperands` plus `FromNamedDoesNotSuppressTheDefaultGraphClause` — the latter is your case.

Trinity/Stores/StoreBase.cs
meberl: Right on both counts, fixed in 01d3b6b.

Rather than add the guard to Virtuoso, I took the other option you offered: every store's `ContainsModel(IModel)` override was a verbatim delegation to the `Uri` overload, so Virtuoso's and GraphDB's are deleted and `StoreBase` is now the single implementation. That is what makes the guard actually reach them — centralizing it while leaving the overrides in place fixed nothing, which was your point.

The test moved into a shared `StoreCatalogTest<T>`, derived by Fuseki, GraphDB, Virtuoso and the in-memory store. **It immediately found the same defect class in two more backends**, neither caused by this branch:

- **Virtuoso `ContainsModel(Uri)` returned `true` for every URI.** It ran `SPARQL ASK { GRAPH <g> { ?s ?p ?o } }` and tested `result.Rows.Count > 0` — but an `ASK` always returns exactly one row, the row *holding* the boolean, so the count was never zero and the answer was never read. Identical shape to the Fuseki defect: store consulted, answer discarded. Now a `SELECT … LIMIT 1`, for which the row count is the question.
- **The in-memory store answered `true` for a null URI** — `GraphName(null)` yields `null` and dotNetRDF reads that as the default graph, which always exists. Guarded, so all four agree.

So `ContainsModel` had a test the whole time, on the one backend whose implementation was trivially correct. Your ADR correction was too generous to Virtuoso — it was wrong for the `IModel` overload *and* the `Uri` one; the ADR now says so.

Trinity.Fuseki/FusekiStore.cs
meberl: Confirmed and fixed in 01d3b6b, with your suggested guard.

Verified against a live 5.1.0 server holding one IRI-named and one bnode-named graph: `ListGraphNames()` returns `[b0, http://ex/g1]` while the obsolete `ListGraphs()` returns only the IRI, and `ListModels()` threw `UriFormatException: Invalid URI` — so the enumeration died part-way rather than skipping the one graph nobody can name. Test added.

On the shared-helper question: Fuseki is the only backend exposed today. Virtuoso builds its list from a `SELECT DISTINCT ?g` query and GraphDB overrides `ListGraphs()` in its own connector, so neither goes through `ListGraphNames()`. I have recorded the shared-helper idea in ADR-0043's follow-ups rather than building it for a single caller — say the word if you would rather have it now.

Trinity/Query/SparqlPreprocessor.cs
meberl: Verified in 01d3b6b — resolving.

`DefaultGraphs` and the new `NamedGraphs` are separate sets, `AddGraph` de-dupes against the one matching the clause it is inserting, and `DeclaredGraphsForCurrentClause` routes the recording side by `NAMED`/`FROMNAMED`. My repro now emits both clauses exactly once. `FromNamedDoesNotSuppressTheDefaultGraphClause` pins that case and `DeclaredDefaultGraphsExcludeFromNamedOperands` pins the `GetDefaultModels()` contract; both pass.

The `Bind` symptom falls out of the same split — a `FROM @model` parameter is a default-graph clause, so a `FROM NAMED` operand in the query text no longer collides with it. Good catch on `GraphDBStore.cs:269` re-emitting `GetDefaultModels()` as `FROM`; keeping that accessor `FROM`-only is the right call given how both callers read it.

One thing left as-is, not a blocker: `Bind` still records `GRAPHPARAMETER` values into `DefaultGraphs` unconditionally, and `GRAPHPARAMETER` is assigned after `Token.FROMNAMED` as well as `Token.FROM` (`SparqlPreprocessor.cs:254-257`). So `FROM NAMED @g` would bind into the wrong set — the same conflation, on the parameter path rather than the literal-URI path. Trinity emits no such query today, so it is unreachable; worth an ADR follow-up line rather than code.

Trinity/Stores/StoreBase.cs
meberl: Verified in 01d3b6b — resolving.

Deleting the overrides rather than duplicating the guard is the better of the two options; `StoreBase.ContainsModel(IModel)` is now the single implementation and `grep` confirms only the `Uri` overload is overridden anywhere (`SparqlEndpointStore` still declares its own pair, but it implements `IStore` directly and both throw `NotSupportedException`, so it is unaffected).

Both follow-on defects check out, and finding them is a good argument for the shared fixture:

- Virtuoso `ContainsModel(Uri)` is now `SELECT ?s … LIMIT 1` — your reading of the `ASK` bug is right, the single row holding the boolean made `Rows.Count > 0` a constant.
- `dotNetRDFStore.ContainsModel(Uri)` guards null before `HasGraph(GraphName(uri))`.

`StoreCatalogTest<T>` is derived by all four backends. Ran it live: `DotNetRDFCatalogTest` 2/2, `VirtuosoCatalogTest` 2/2 against a self-provisioned Virtuoso container, `FusekiCatalogTest` 2/2 against Fuseki 5.1.0. Correction accepted on the ADR — the `Uri` overload was broken too, so my original wording was indeed too generous.

Trinity.Fuseki/FusekiStore.cs
meberl: Verified in 01d3b6b — resolving.

The guard is in and the comment records why `ListGraphNames()` differs from the obsolete `ListGraphs()`, which is the part a future reader would otherwise re-discover the hard way. Fuseki suite passes against 5.1.0.

Agreed on deferring the shared helper — I checked your reasoning and it holds: Virtuoso builds its list from `SELECT DISTINCT ?g` and GraphDB overrides `ListGraphs()` in its own connector, so `ListGraphNames()` genuinely has one caller today. A helper for one caller is speculative; the ADR follow-up line is the right weight. No need to build it now.

PR #43, opened 2026-08-24
Fix the store suites: eight inferencing failures were provisioning gaps, not store limitations

meberl:
## Eight failing tests that were never store limitations

Virtuoso and GraphDB each carried four failing inferencing tests. They had been red long enough to read
as settled facts about the backends. Both were **provisioning gaps** — the same category as the missing
Fuseki dataset in #42.

Confirmed pre-existing before touching anything, by running both suites against a clean `develop` in a
worktree: identical counts, identical test names.

**Virtuoso never compiled the query:**

```
SP031: SPARQL compiler: 'define input:inference refers to
undefined inference rule set "urn:semiodesk/test/ruleset"
```

`VirtuosoContainer` puts that rule-set name into the connection string, but nothing creates it. It was
declared in `tests/Trinity.Tests.Virtuoso/ontologies.config` and built by the configuration subsystem
**ADR-0011 retired**. That file is still in the repo and nothing reads it — a dead declaration with a
live reference to it.

**GraphDB compiled the query and returned nothing.** Its reasoner *is* configured
(`graphdb:ruleset "rdfsplus-optimized"`), but the axiom it would reason over was never loaded.
`GetTypedResourcesWithInferencingTest` needs `nco:PersonContact rdfs:subClassOf nco:Contact`, and the
shared `TestOntologies` seeds rdf, rdfs, owl, foaf and the space-test ontology — **not nco** — even
though `nco.trig` sits in the same folder and the per-store setups it replaced all loaded it.

ADR-0043 recorded GraphDB's failures as "an unresolved issue there". That was wrong, and this PR
corrects it.

## Results

| suite | before | after |
|---|---|---|
| **Virtuoso** | 236 passed / **4 failed** | **240 / 0** |
| **GraphDB** | 243 passed / **4 failed** | **247 / 0** |
| Fuseki | 249 / 0 | 249 / 0 |
| in-memory | 607 / 3 | 607 / 3 (pre-existing `UriRef`) |
| generator / vocabulary | 23 / 29 | 23 / 29 |

**Every store backend is now at zero failures.**

## The subtle part

`rdfs_rule_set` builds the rule set from a **snapshot of the graphs taken when it runs**. Registering it
at container start — before anything is seeded — produces a rule set with no rules, which compiles fine
and infers nothing. Measured on a live server:

| when `rdfs_rule_set` runs | entailed row returned |
|---|---|
| before the data is loaded | no |
| re-registered, same name, after the data | **yes** |
| fresh name, after the data | **yes** |

Hence a new `IStoreTestSetup.AfterSeed(IStore)` hook with a **default no-op body**, so only Virtuoso
implements it and the other setups are untouched. The start-up registration is kept as well, so the name
resolves and a query naming it compiles even if seeding fails — one clear error instead of a silently
empty result.

## One shared test changed — worth a second opinion

Virtuoso's last failure was `Expected: 3, But was: 8` in `SparqlQueryTest.TestInferencing`. The assertion
counts contact media, but under inferencing the type is *entailed*, and SPARQL `SELECT` is a **bag**: a
store that back-chains (Virtuoso) yields one solution per derivation path where one that materializes
(GraphDB) yields one per subject. Neither is wrong — the query has to say which it means, so it now says
`SELECT DISTINCT`. The three media have distinct dates, so `DISTINCT` collapses only the duplicates.

I believe this is a genuine test bug rather than a fudge, but it is a shared assertion changed to
accommodate one backend, so it is the judgement call here most worth review.

## Inferencing was effectively untested until now

No suite anywhere exercised a working reasoner: the in-memory store does not implement inferencing,
Fuseki cannot switch it per query (#42), and both remaining backends were misconfigured. `inferenceEnabled`
was, in effect, untested everywhere.

## CI — a separate commit, easy to drop

`906ce7c` runs the store suites in CI. ADR-0036 excluded them for "Docker availability + large image
pulls" and anticipated a separate job later; both objections are answered — GitHub-hosted Linux runners
ship Docker, and **this repository is public, so standard runners cost nothing**. The third, unstated
objection, that eight tests were red, stops holding with the first commit.

A matrix over the three backends, so each pulls only its own image, they run in parallel, and a failure
names the store in the check title. `fail-fast: false`, because the useful question is *which* backends
a change breaks. `.trx` files upload on failure. The existing `build` job is untouched and still runs
only the in-memory suites, so PR feedback stays fast and a Docker hiccup cannot redden it.

Costs roughly 2 GB of image pulls per run — free on a public repo, a few minutes of wall clock.

## Also in here

`498b54d` reverses an over-correction of mine from `27259f4`. Told that dotNetRDF has inferencing and
that 3.x did not remove it, I swung from "the in-memory store has none" to wording that implied the
feature works because the reasoner classes exist. Both framings were wrong, and the repo already had the
right one in `doc/known-test-failures.md`: the in-memory store **does not implement inferencing**.
Half-wired parts are not support. The four `Assert.Inconclusive` messages now say that plainly instead of
surveying what dotNetRDF could do in principle.

## Verifying locally

Needs a Docker daemon.

```bash
dotnet test tests/Trinity.Tests.Virtuoso/Trinity.Tests.Virtuoso.csproj -c Release
dotnet test tests/Trinity.Tests.GraphDB/Trinity.Tests.GraphDB.csproj -c Release
dotnet test tests/Trinity.Tests.Fuseki/Trinity.Tests.Fuseki.csproj -c Release
```

## Follow-up already on a branch

`feature/in-memory-inferencing` builds on this one and finishes in-memory inferencing (ADR-0045), taking
that suite to 620 passed and bringing seven tests off the bench. It depends on the nco seeding here, so
this should land first.


Trinity.Tests/TestOntologies.cs
```diff
@@ -51,6 +51,12 @@ public static void LoadInto(IStore store)
                 (new Uri("http://www.w3.org/2002/07/owl#"),              "Ontologies/owl.n3"),
                 (new Uri("http://xmlns.com/foaf/0.1/"),                  "Ontologies/foaf.rdf"),
                 (new Uri("http://www.example.com/myontology"),           "Ontologies/space test ontology.ttl"),
+                // nco carries the class hierarchy the inferencing tests reason over -- notably
+                // nco:PersonContact rdfs:subClassOf nco:Contact. The per-store setups this replaced
+                // all loaded it; the shared version did not, so every backend with a working
+                // reasoner still answered 0 and the tests read as "inferencing is broken here".
+                (new Uri("http://www.semanticdesktop.org/ontologies/2007/03/22/nco#"),
```
meberl: **Seeding this file makes a documented follow-up into a live bug (Fuseki).**

`nco.trig` holds **two** named graphs — `…/nco#` at line 22 and `…/nco_metadata#` at line 794. In `FusekiStore.cs:442` the `DeleteGraph` sits *inside* the per-graph loop and targets the caller's `graphUri`, not `g.Name`:

```csharp
foreach (Graph g in s.Graphs)
{
    if (!update && exists)
    {
        Connector.DeleteGraph(graphUri);   // <-- always the caller's URI
    }

    Connector.SaveGraph(g);
}
```

With `exists = Connector.HasGraph(graphUri)` (`FusekiStore.cs:416`), the first fixture in a fresh container is fine — `exists` is false, nothing is deleted, both graphs land. From the **second** fixture onward `exists` is true, and the iteration saving `nco_metadata#` deletes the `nco#` graph the previous iteration just wrote. So the axioms this PR adds to fix inferencing get removed again, for every fixture after the first.

ADR-0043 lists this delete-in-loop as a deliberate follow-up, which was defensible while nothing seeded a multi-graph TriG file. This PR seeds one, so it becomes this PR's problem. The fix is to hoist the delete out of the loop and key it on the graph actually being written:

```csharp
var seen = new HashSet<string>();

foreach (Graph g in s.Graphs)
{
    if (!update && seen.Add(g.Name?.ToString()) && Connector.HasGraph(g.Name))
    {
        Connector.DeleteGraph(g.Name);
    }

    Connector.SaveGraph(g);
}
```

Worth a test that seeds twice and asserts both graphs survive — that is the shape that would have caught it.

Trinity.Tests/TestOntologies.cs
```diff
@@ -51,6 +51,12 @@ public static void LoadInto(IStore store)
...
+                                                                         "Ontologies/nco.trig"),
```
meberl: **Same delete-in-loop defect on GraphDB, with an extra wrinkle that affects this PR's headline result.**

`GraphDBStore.cs:439` has the identical shape — `DeleteGraph(graphUri)` inside `foreach (Graph g in s.Graphs)` — so the `nco#`/`nco_metadata#` pair wipes the same way as on Fuseki.

The wrinkle: GraphDB's `exists` is computed from the **obsolete `ListGraphs()`**, not `HasGraph`. So exactly one of these is true:

- `ListGraphs()` reports the graph, `exists` is true, and the second iteration wipes `nco#`; or
- `ListGraphs()` misses it, `exists` is always false, and `update: false` never actually replaces a graph — a different latent bug.

Either way something is wrong, and **which one it is decides whether GraphDB's 4→0 inferencing result is real** or an artifact of the seed order. Since that number is the claim in ADR-0044, it is worth pinning down rather than inferring. Logging `exists` for the `nco` seed on GraphDB once would settle it in a single run.

tests/Trinity.Tests.Virtuoso/VirtuosoContainer.cs
```diff
@@ -64,8 +88,65 @@ public async Task StartAsync()
...
+
+        private static async Task CreateRuleSetAsync(IContainer container)
+        {
+            var statements = string.Concat(
+                RuleSetGraphs.Select(g => $"rdfs_rule_set('{RuleSet}', '{g}');"));
+
+            var result = await container.ExecAsync(
+                new[] { "isql", "1111", "dba", "dba", $"exec={statements}" });
+
+            if (result.ExitCode != 0)
```
meberl: **This guard cannot detect the failure it exists to prevent.**

It checks only `result.ExitCode`, but `isql exec=` reports SQL errors on **stdout** and still exits 0. And `rdfs_rule_set` over a graph that does not exist is not an error at all — it registers the name with no axioms behind it.

That is precisely the state this PR diagnoses: the rule set resolves, queries compile, and every `inferenceEnabled` query silently returns zero rows. The guard would pass. Note the comment at line 99 already says the first call "cannot yet infer anything" — so the empty-rule-set state is known-reachable by design here; nothing distinguishes it from the broken kind.

Stronger check, in rough order of value:

1. After `RefreshRuleSet()`, assert an actual entailment — run the `nco:PersonContact`/`nco:Contact` query with `inferenceEnabled: true` and require a non-empty result. That is the property the suite depends on, and it is cheap because the axioms are already loaded at that point.
2. Failing that, at minimum scan `result.Stdout` for Virtuoso's error markers (`*** Error`, `SQLState`) and throw on a hit.

Option 1 is what turns "the name resolves" into "inference works", which is the difference this PR is about.

tests/Trinity.Tests.Virtuoso/VirtuosoContainer.cs
```diff
@@ -45,6 +47,28 @@ namespace Semiodesk.Trinity.Tests.Virtuoso
...
+        /// pass <c>inferenceEnabled: true</c> depend on.
+        /// </summary>
+        private const string RuleSet = "urn:semiodesk/test/ruleset";
+
+        /// <summary>
+        /// Schema graphs the rule set draws its axioms from. These are the graphs
+        /// <see cref="TestOntologies"/> seeds, and the same set the retired
+        /// <c>ontologies.config</c> declared for this rule set.
+        /// </summary>
+        private static readonly string[] RuleSetGraphs =
```
meberl: **The doc comment's parity claim is already false.**

The comment says "These are the graphs `TestOntologies` seeds", but this array has five entries and omits `http://www.example.com/myontology`, which `TestOntologies` does seed (`TestOntologies.cs:53`).

The list itself is not wrong — the five do match the retired `ontologies.config` for this rule set exactly, which I verified, and that is a reasonable thing to preserve. It is the *claim* that is wrong, and it matters more than a usual comment nit: this array and the seed list are hand-maintained in two files, and drift between them degrades inferencing to zero rows **silently** — the exact failure this PR spent its effort diagnosing.

Two options:

- Derive the array from `TestOntologies`' graph list, so drift is impossible, and note the deliberate exclusion (if `myontology` is intentionally out, say why — it has no axioms worth reasoning over).
- Or keep it hand-written but fix the comment to say "the graphs the retired `ontologies.config` declared for this rule set", which is what it actually is, and drop the `TestOntologies` parity claim.

The first is better given the silent-failure mode.

meberl (commented):
Review of the store-suite inferencing fix. 16 files, test infrastructure + docs, no runtime source changes.

**The premise is right and well-evidenced.** Virtuoso's rule set really was declared only in the `ontologies.config` ADR-0011 retired, and the shared `TestOntologies` really never seeded `nco.trig` — so both were provisioning gaps, not store limitations. `AfterSeed` is the right seam. The note that `rdfs_rule_set` builds from a *snapshot* taken when it runs, hence create-then-refresh, is a genuinely non-obvious detail worth having in the file.

Seven findings. Three of them (the two delete-in-loop comments and the `ExitCode` guard) share the failure mode this PR is about: **silently returning zero rows**. Those are what I'd want addressed before merge.

One note on placement: the first two findings are in `Trinity.Fuseki/FusekiStore.cs` and `Trinity.GraphDB/GraphDBStore.cs`, which are not in this diff, so I anchored them on the `nco.trig` seeding line that makes them reachable.

Checked and cleared: all thirteen shared fixtures derive from `StoreTest<T>`, so `AfterSeed` runs after every seed; the default interface method is fine on net8.0 and `VirtuosoTestSetup` implements it through an interface-typed reference; `rdfs_rule_set` is idempotent by name, so the refresh cannot duplicate rules; the three `dc:date` values are distinct, so the added `SELECT DISTINCT` cannot drop the count below 3; no test asserts on model counts, so the extra `nco` graph is inert elsewhere; the five rule-set graphs do match the retired `ontologies.config` exactly.

tests/Trinity.Tests.Virtuoso/VirtuosoContainer.cs
```diff
@@ -64,8 +88,65 @@ public async Task StartAsync()
...
+            // ADR-0011 retired -- that file is still in the repo but nothing reads it -- so every
+            // inferenceEnabled query failed with "SP031: 'define input:inference refers to undefined
+            // inference rule set". Create it here, the way GraphDBContainer creates its repository.
+            //
+            // This first call only makes the name resolve, so a query naming it compiles. It cannot
+            // yet infer anything: rdfs_rule_set builds the rule set from a *snapshot* of the graphs
+            // taken when it runs, and nothing is seeded yet. RefreshRuleSet below rebuilds it once
+            // TestOntologies has loaded the axioms -- measured: created-before-data infers nothing,
+            // re-registering the same name after the data returns the entailed row.
+            await CreateRuleSetAsync();
```
meberl, line 112: **No retry around the first `isql` call.**

The readiness strategy is `UntilMessageIsLogged("Server online at 1111")` (line 86), which tells you the listener printed a banner — not that it accepts connections. `CreateRuleSetAsync()` then runs `isql` immediately with no retry, and a failure here throws from `[OneTimeSetUp]`, taking the **whole assembly** down rather than one test.

That is a flake shape that gets much more expensive now that this runs on every PR in the `stores` matrix, and it will read as "Virtuoso is broken" rather than "the container was slow".

A short bounded retry (a few attempts, brief backoff) around the `ExecAsync` would cover it. Alternatively make the wait strategy an actual connection probe rather than a log match — worth it for a container that boots as slowly as Virtuoso.

.github/workflows/ci.yml
```diff
@@ -74,3 +74,48 @@ jobs:
...
+            8.0.x
+            10.0.x
+
+      - name: Restore
+        run: dotnet restore tests/Trinity.Tests.${{ matrix.store }}/Trinity.Tests.${{ matrix.store }}.csproj
+
+      - name: Build
+        run: dotnet build tests/Trinity.Tests.${{ matrix.store }}/Trinity.Tests.${{ matrix.store }}.csproj -c Release --no-restore
+
+      # Testcontainers pulls the pinned image and binds the service port to a random free host port,
```
meberl: **"the pinned image" is not true for Virtuoso.**

GraphDB is pinned (`ontotext/graphdb:10.8.0`) and Fuseki is pinned (`stain/jena-fuseki:5.1.0`), but `VirtuosoContainer.cs:81` uses `openlink/virtuoso-opensource-7:latest`.

That was low-stakes while these suites were opt-in and local. Now that this job runs on every PR, an upstream push to `:latest` can redden the matrix on a branch that changed nothing — and, given ADR-0043's note that Jena 4.x vs 5.x changes `urn:uuid:` behaviour, a silent major bump is exactly the kind of thing that costs an afternoon.

Pin Virtuoso to the digest or tag you actually validated against, the way the other two are. Then this comment becomes accurate as written.

CLAUDE.md
```diff
@@ -69,9 +69,13 @@ dotnet pack Trinity/Trinity.csproj -c Release          # -> Semiodesk.Trinity.2.
 - Store integration tests (`tests/Trinity.Tests.*`) **self-provision** their server in Docker via
   Testcontainers on a random host port (ADR-0036): run `dotnet test tests/Trinity.Tests.{Virtuoso,GraphDB,Fuseki}`
   with a Docker daemon running. Excluded from the default CI job (Docker + large images). Current:
```
meberl: **This bullet now contradicts itself.**

Line 71 still says the store suites are "Excluded from the default CI job (Docker + large images)", and line 73 — two sentences later — says "They also **run in CI** now, as a separate `stores` matrix job". Both readings are defensible in isolation (a separate job is not the default job), but read straight through it looks like one of the two is stale, and a reader cannot tell which.

Two smaller things in the same paragraph:

- Line 78 splices the new text mid-sentence into the old: "…(ADR-0022). Fuseki's long-standing \"4/86, blocked on an upstream `FusekiConnector` bug\" was a **misdiagnosis**…" — the Fuseki subject changes meaning between the two halves.
- The in-memory counts in the Build & test block above still read `602 passed, 7 skipped`, where ADR-0044 says 607/3. I measured 607 passed / 3 failed / 7 skipped on `feature/fuseki-revival`, so 602/7 is stale.

Suggest rewriting the bullet so the CI status is stated once: they self-provision via Testcontainers, they run in CI as the `stores` matrix job (ADR-0044), and the ADR-0036 exclusion no longer applies — with the reason it no longer applies, which lines 81-83 of the workflow already explain well.

Trinity.Tests/TestOntologies.cs
meberl: Confirmed and fixed in `9618e67` — and measuring it found a second, larger half you did not predict.

The delete-in-loop is real exactly as you describe. But `BaseUri` was also never assigned per graph, and the connector derives its target from it (ADR-0038), so **every** graph went to the *default* graph and the last one overwrote the rest. Measured against a live Fuseki 5.1.0 before the fix:

```
graphs on the server: (none)
default graph:        12 triples     <- nco_metadata#, having overwritten nco#'s 542
```

So the axioms were not being wiped by the delete — the delete never even fired, because `exists = HasGraph(nco#)` was always false, `nco#` having never been created. They were lost to the overwrite. The non-TriG branch had assigned `BaseUri` for precisely this reason all along.

Both halves fixed: delete the graph being written, and set its `BaseUri` first. After the fix, across three consecutive re-reads on both backends:

```
after seed #1:  nco# = 542   nco_metadata# = 12   PersonContact-subClassOf: True
after seed #2:  nco# = 542   nco_metadata# = 12   PersonContact-subClassOf: True
after seed #3:  nco# = 542   nco_metadata# = 12   PersonContact-subClassOf: True
```

Your suggested test shape is now `MultiGraphTrigTest<T>`, shared and derived by Fuseki and GraphDB: it asserts both graphs are populated, re-reads the file the way a second fixture's seeding does, and asserts neither disappeared.

One correction to the finding as written: the two graph headers are at lines 794 (`nco_metadata#`) and — implicitly — the file's own `@base`, not line 22, which is inside the licence comment block. Does not change the conclusion.

Trinity.Tests/TestOntologies.cs
meberl: Fixed in `9618e67`, and the dilemma resolves to your **second horn** — settled by measurement rather than inference, as you asked.

`ListGraphs()` never reported `nco#`, because `nco#` was never created as a named graph (see the reply above: everything went to the default graph). So `exists` was always false, the delete never fired, and `update: false` never actually replaced anything — the latent bug you named.

**GraphDB's 4→0 is real.** Queried directly against a live repository after seeding:

```
SELECT ?o WHERE { nco:PersonContact rdfs:subClassOf ?o }
  -> rdfs:Resource, nco:Contact, nco:PersonContact, nco:Role
```

The axioms reached the repository's default graph, and GraphDB reasons repository-wide, so the inferencing tests genuinely pass. Default graph held 752 triples against 554 asserted — the difference being GraphDB's own materialized entailments. Not an artifact of seed order.

After the fix the graphs land under their own names and `nco#` is queryable via `FROM <nco#>`; GraphDB stays at 0 failures (248 now, +1 from the new test). Recorded in ADR-0044 so the next reader does not have to re-derive it.

tests/Trinity.Tests.Virtuoso/VirtuosoContainer.cs
meberl: Right, and this was the sharpest of the seven. Fixed in `9618e67` with your option 1.

`RefreshRuleSet()` now asserts an actual entailment rather than that the name resolves — `nco:PersonContact` must come back with `nco:Contact` among its superclasses, which the seeded nco schema guarantees. That is the property the suite depends on, and as you say it is cheap at that point because the axioms are already loaded.

Your option 2 is in as well, since it is nearly free and catches a different class of failure: `isql exec=` reports SQL errors on stdout and still exits 0, so the output is now scanned for `*** Error` / `SQLState` and the exit code alone is never trusted.

You were also right that the empty-rule-set state is reachable *by design* here — the start-up registration deliberately creates a rule set with no axioms behind it so the name resolves. That is exactly why the check had to move to `RefreshRuleSet`, after seeding, where "it works" is a checkable claim instead of an assumption.

tests/Trinity.Tests.Virtuoso/VirtuosoContainer.cs
meberl: Fixed in `9618e67`, taking your first option.

`RuleSetGraphs` is now derived from `TestOntologies.Graphs` rather than restated, so drift is impossible:

```csharp
private static IEnumerable<string> RuleSetGraphs =>
    TestOntologies.Graphs.Select(g => g.Graph.OriginalString);
```

That required making `TestOntologies` public and lifting its seed list into a `Graphs` field, which `LoadInto` now consumes — one list, one source of truth.

`myontology` is therefore in the rule set now rather than deliberately excluded. It contributes no axioms worth reasoning over, so this is inert, and I preferred that to encoding an exclusion that would itself need maintaining. The reason the old five matched `ontologies.config` exactly is simply that `ontologies.config` predates `myontology` being seeded.

Agreed on why this one mattered more than a comment nit: two hand-maintained copies of a list whose divergence degrades inferencing to zero rows silently is the same failure shape as everything else in this PR.

tests/Trinity.Tests.Virtuoso/VirtuosoContainer.cs
meberl: Fixed in `9618e67`. Bounded retry — five attempts, 2s apart — around every `isql` call, not just the first.

Your reasoning about the cost is what decided it: a log-line match says the banner printed, and a failure in `[OneTimeSetUp]` takes down the whole assembly rather than one test, so this reads as "Virtuoso is broken" on a branch that changed nothing. That was tolerable when the suite was opt-in and local; it is not now that it runs on every PR.

I kept the log-line wait strategy and layered the retry on top rather than replacing it with a connection probe. The retry covers the same window and is contained in one place; if Virtuoso's boot time turns out to be flaky in practice on CI runners, a real probe is the next step rather than a longer backoff.

.github/workflows/ci.yml
meberl: Fixed in `9618e67` — pinned to `openlink/virtuoso-opensource-7:7.2.14`, so the comment is now accurate as written.

Worth noting the pin was not a no-op: `:latest` and `7.2.14` are *different images* (`0dbe1ab4fa0c` vs `deff8b5bad73`), so I re-ran the suite on the pinned tag rather than assuming equivalence. Still 240 passed / 0 failed.

Your ADR-0043 point is the one that makes this worth doing rather than a tidy-up — a silent major bump is precisely how the Jena 4.x/5.x `urn:uuid:` behaviour change would have arrived, and that one cost real time to diagnose.

CLAUDE.md
meberl: All three fixed in `9618e67`.

The contradiction was the worst of them: a reader cannot tell which of two adjacent sentences is stale, and the stale-looking one is the true one. The bullet now states the CI status once — they self-provision via Testcontainers, they run as the `stores` matrix job, and the ADR-0036 exclusion no longer applies, with the reason.

The mid-sentence splice is gone; the Fuseki misdiagnosis is its own paragraph rather than grafted onto the end of the inferencing one.

On the counts: you are right that 602/7 was stale, and it is now 607. It will move again immediately — the follow-up branch takes the in-memory suite to 620 as the inferencing tests come off the bench — but leaving a number that was wrong on this branch to be fixed by a later one is how it got stale in the first place.

Trinity.Fuseki/FusekiStore.cs
```diff
@@ -435,11 +435,31 @@ public override Uri Read(Uri graphUri, Uri url, RdfSerializationFormat format, b
...
+                        //  - BaseUri has to be set per graph. The connector derives its target from it
+                        //    (ADR-0038), so leaving it null sent every graph to the *default* graph,
+                        //    where the last one silently overwrote the rest. Measured on nco.trig:
+                        //    542 triples written, then replaced by the 12 of nco_metadata#.
+                        //
+                        // The non-TriG branch below already assigns BaseUri for exactly this reason.
                         foreach (Graph g in s.Graphs)
                         {
-                            if (!update && exists)
+                            if (!(g.Name is IUriNode graphName))
```
meberl: **New: the fix silently drops a TriG file's default-graph triples.**

The `continue` skips any graph whose name is not an `IUriNode` — which includes the file's **default graph**. Before this change those triples were passed to `SaveGraph(g)` and landed somewhere (the connector's default graph); now they are discarded, and `Read` still returns `graphUri` either way, so a caller cannot tell that part of the file was ignored.

For `nco.trig` this is moot — see below — but for any TriG that mixes default-graph triples with named ones, `Read` now silently loses data. Given that the whole PR is about silent zero-row failures, this one deserves a decision rather than a fall-through: either write unnamed triples to `graphUri` (the caller *did* name a target graph, which is a reasonable home for them), or keep skipping but say so out loud. The comment above explains why `graphUri` is not each graph's name; it does not say what happens to triples that have no name at all.

---

**Correction to the record, in both directions.** Your reply said the two graph headers are at line 794 and "implicitly, the file's own `@base`, not line 22, which is inside the licence comment block." You are right that my line 22 was wrong — I misread a filtered `grep -n` (22nd non-comment line) as a file line. But the alternative is also wrong: `nco.trig` has **two explicitly named graphs**, not one named plus an implicit default.

```
55: @prefix nco:  <http://www.semanticdesktop.org/ontologies/2007/03/22/nco#> .
58: nco: {nco:region          <- named graph, prefixed name
794: <http://...nco_metadata#> {  <- named graph, absolute IRI
```

This is worth correcting because **the fix depends on it**. If the 542-triple block really were the unnamed default graph, the `continue` on this line would skip it outright and the fix would drop the class hierarchy it exists to preserve. It works precisely because that graph *is* named. As written, the code is correct and the reasoning recorded next to it is not — which is the combination that gets "simplified" into a bug later. Worth fixing in ADR-0044 too.

Trinity.GraphDB/GraphDBStore.cs
```diff
@@ -432,11 +432,22 @@ public override Uri Read(Uri graphUri, Uri url, RdfSerializationFormat format, b
...
                         store.LoadFromFile(path, new TriGParser());
 
+                        // See the matching comment in FusekiStore.Read: graphUri names the file, not
+                        // the graphs inside it. The delete has to target the graph being written, and
+                        // BaseUri has to be set per graph or the connector sends them all to the
+                        // default graph (ADR-0038).
                         foreach (var g in store.Graphs)
                         {
-                            if (!update && exists)
+                            if (!(g.Name is IUriNode graphName))
```
meberl: Same unnamed-graph drop as the Fuseki side — whatever is decided there should land here too, since this branch deliberately mirrors it.

Separately, thank you for settling the `ListGraphs()` dilemma by measurement instead of inference. The answer being the second horn — `exists` always false, so `update: false` never replaced anything — is the more useful outcome, and the direct query showing `nco:PersonContact` entailing `nco:Contact`, plus 752 materialized against 554 asserted, is what makes "GraphDB's 4→0 is real" a checked claim rather than a hoped-for one. Confirmed on my side: GraphDB runs **249 passed / 0 failed / 1 skipped**.

meberl (commented):
Rechecked at `1885fc3`. **All seven findings are genuinely fixed**, and I verified the results rather than taking the replies on report — built clean, then ran every suite:

| suite | ADR-0044 says | I measured |
|---|---|---|
| Virtuoso | 240 passed / 0 failed | **241 passed / 0 failed / 1 skipped** |
| GraphDB | 247 passed / 0 failed | **249 passed / 0 failed / 1 skipped** |
| Fuseki | 249 passed / 0 failed | **251 passed / 0 failed / 1 skipped** |
| in-memory (.NET 10) | 607 passed / 3 failed | **608 passed / 3 failed / 7 skipped** |
| in-memory (.NET 9) | — | **611 passed / 0 failed / 7 skipped** |

Every backend is green. Spot-checks that mattered: the `MultiGraphTrigTest<T>` re-read assertion is the right shape and passes on both backends; `RuleSetGraphs` is now derived from `TestOntologies.Graphs`, so drift is structurally impossible; `RefreshRuleSet` asserts a real entailment (`nco#Contact` in the result) rather than that the name resolves; the retry is 5 × 2 s around *every* `isql` call; the image is pinned to `7.2.14`. The `.trx` files added in `9618e67` were removed again in `1885fc3`, so the net diff is clean.

I also confirmed `known-test-failures.md` is right about the three URI-equality tests, having briefly doubted it: `DOTNET_ROLL_FORWARD=Major` resolves to .NET **9** here and all three pass; `LatestMajor` resolves to .NET **10** and all three fail. Runtime-dependent exactly as documented, not order-dependent. CI pins net8.0, so it is unreachable there.

Two follow-ups below — one is a real behaviour change the fix introduced, the other is the counts going stale again. Neither is large.

On `e106053` (not part of my original review): making the w3.org fetch inconclusive rather than failing is the right call, and splitting it into its own test is better than burying it. I checked the one thing that could have weakened it — `StoreTest.SetUp` clears `Model1`, so `Assert.IsFalse(Model1.IsEmpty)` still means what it says.

doc/adr/0044-store-suites-green-and-in-ci.md
```diff
@@ -0,0 +1,145 @@
...
+## Consequences
+
+- **The whole suite is green**, on every backend:
+
+  | suite | before | after |
+  |---|---|---|
+  | Virtuoso | 236 passed / 4 failed | **240 passed / 0 failed** |
+  | GraphDB | 243 passed / 4 failed | **247 passed / 0 failed** |
+  | Fuseki | 249 passed / 0 failed | 249 passed / 0 failed |
+  | in-memory | 607 passed / 3 failed | 607 passed / 3 failed |
```
meberl: **The results table went stale inside this PR.** Every row is now low, because `9618e67` added `MultiGraphTrigTest` and `e106053` added `ReadFromUrlTest` after the table was written:

| row | table | measured at `1885fc3` |
|---|---|---|
| Virtuoso | 240 / 0 | 241 / 0 (+1 skipped) |
| GraphDB | 247 / 0 | 249 / 0 (+1 skipped) |
| Fuseki | 249 / 0 | 251 / 0 (+1 skipped) |
| in-memory | 607 / 3 | 608 / 3 on .NET 10; 611 / 0 on .NET 9 |

The conclusion the table supports — all backends green — is unchanged and correct. It is the numbers that drifted.

Two suggestions, since this is the second time counts have gone stale in as many PRs:

- The rows also omit the skipped column, which is why `249` reads as the total for Fuseki when the total is 252. Adding it would make the next drift self-evident.
- The in-memory row is worth annotating with the runtime, not just the footnote below it. `608 / 3` and `611 / 0` are both true and differ only by which runtime the net8.0 assembly rolls forward onto — I had to measure both to work out that the footnote was right.

CLAUDE.md
```diff
@@ -55,7 +55,7 @@ is netstandard2.0 / net8.0 and builds cross-platform.
 
 ```bash
 dotnet build Semiodesk.Trinity.sln -c Release          # whole solution, SDK-only
-dotnet test Trinity.Tests/Trinity.Tests.csproj         # 602 passed, 7 skipped (quarantined)
+dotnet test Trinity.Tests/Trinity.Tests.csproj         # 607 passed, 7 skipped (quarantined)
```
meberl: Same drift here: `607 passed, 7 skipped` is now `608 passed, 3 failed, 7 skipped` on .NET 10, or `611 passed, 0 failed, 7 skipped` on .NET 9 — `e106053` added a test after this line was updated.

The rest of the CI paragraph reads much better now. Stating the status once, with the reason the ADR-0036 exclusion no longer applies, is exactly what was missing, and moving the Fuseki misdiagnosis into its own paragraph fixed the subject change mid-sentence.

One thought on the count itself, given your note that the follow-up branch takes this to 620: a bare number in a command comment is guaranteed to go stale and nothing checks it. Either drop it, or say what it is for — if the point is "a non-zero failure count here is a regression", the skipped/quarantined count is the number worth pinning, and it has been stable at 7.

Trinity.Fuseki/FusekiStore.cs
meberl: Both parts right. Fixed in `5b82664`.

**The drop.** You are correct that this was the same silent-data-loss shape arriving inside the fix for it, and it deserved a decision rather than a fall-through. Unnamed triples now go to `graphUri` — your first option. The caller did name a target, and triples with no graph of their own have no better home; discarding them while returning the same URI gives the caller nothing to notice.

Grouping by target came out of writing it: two graphs can share one target (a file with unnamed triples *and* a graph named `graphUri`), and on a connector whose `SaveGraph` is a PUT the second would replace the first. So graphs are merged per target and written once.

`nco.trig` has no unnamed triples, which is exactly why nothing in the suite would have caught this. `UnnamedTrigTriplesGoToTheGraphTheCallerNamed` writes a file that does, and I verified it fails when the `continue` is put back.

**The correction — you are right and I was wrong.** `nco.trig` has two *explicitly named* graphs. I misread my own grep: `^<http[^>]*>\s*\{` matches `nco_metadata#`'s absolute-IRI label at 794 and misses `nco: {` at 58, because that one is a **prefixed name**. I then reported the absence of a second match as evidence of an implicit default graph, which is reasoning from a tool's blind spot.

And your point about why it matters is the sharper one: had that 542-triple block genuinely been unnamed, the previous commit's `continue` would have dropped the class hierarchy the fix exists to preserve. Code correct, reasoning next to it wrong — the pairing that gets "simplified" into a bug. ADR-0044 now records that both graphs are named, that one uses a prefixed label, and that a grep for absolute IRIs will mislead the next person the same way.

Trinity.GraphDB/GraphDBStore.cs
meberl: Mirrored in `5b82664` — both stores now share a `GroupByTargetGraph` helper, so the unnamed-triple routing and the per-target merge cannot drift between them. GraphDB measures **250 / 0 / 1 = 251** here with the two new TriG tests.

On the `ListGraphs()` dilemma: measuring rather than inferring was your instruction, and it was the right one — I would have guessed the first horn. Worth noting the underlying `ListGraphs()`/`HasGraph` inconsistency is still there; it is now inert, because `exists` is no longer consulted (the delete keys on the graph being written). ADR-0043's follow-up list still carries it as a real cleanup for whoever touches these read paths next.

doc/adr/0044-store-suites-green-and-in-ci.md
meberl: Fixed in `5b82664`, both suggestions taken.

Re-measured every suite on the branch as it now stands rather than adjusting the old numbers:

| suite | before | after |
|---|---|---|
| Virtuoso | 236 / **4** / 1 = 241 | **241 / 0 / 1 = 242** |
| GraphDB | 243 / **4** / 1 = 248 | **250 / 0 / 1 = 251** |
| Fuseki | 249 / 0 / 1 = 250 | 252 / 0 / 1 = 253 |
| in-memory *(.NET 10)* | 607 / 3 / 7 = 617 | 608 / 3 / 7 = 618 |
| in-memory *(.NET 9)* | 610 / 0 / 7 = 617 | **611 / 0 / 7 = 618** |

Totals are included as `passed / failed / skipped = total`, so a row that stops adding up is visible without re-running anything — which is what would have caught both drifts. The in-memory row is split by runtime rather than relying on a footnote, since 608/3 and 611/0 are both true and differ only by roll-forward.

Your numbers and mine agree throughout except GraphDB and Fuseki, where mine are one higher — `5b82664` adds `UnnamedTrigTriplesGoToTheGraphTheCallerNamed` to the shared TriG fixture. Which rather makes your point about how fast these go stale.

CLAUDE.md
meberl: Fixed in `5b82664`, and I took your suggestion rather than just refreshing the number.

The count is now the **quarantine** count, not the passed count:

```
dotnet test Trinity.Tests/Trinity.Tests.csproj   # 7 skipped (quarantined); 0 failed on net8.0
```

You are right that a bare passed count is guaranteed to rot and nothing checks it — it has now gone stale twice in this PR alone. The skipped count is the number that carries meaning: it has been stable at 7, a change to it is a *decision* someone made about a quarantined test, and it is exactly what `doc/known-test-failures.md` documents. A change in the passed count is usually just arithmetic.

I added "on net8.0" because that is the qualifier that makes "0 failed" true — the three `UriRef` failures only appear when the net8.0 assembly rolls forward onto .NET 10, which is the thing you had to measure both runtimes to confirm.

Glad the CI paragraph reads better. Stating it once with the reason was the fix; I had been adding sentences instead of replacing them.

Trinity.Fuseki/FusekiStore.cs
```diff
@@ -483,6 +498,43 @@ public override Uri Read(Uri graphUri, Uri url, RdfSerializationFormat format, b
...
+        /// </summary>
+        /// <remarks>
+        /// A graph carrying its own name is written under that name. Triples carrying none go to
+        /// <paramref name="graphUri"/>, the graph the caller asked to read into: they have no home of
+        /// their own, and silently discarding them would be data loss the caller cannot detect, since
+        /// <c>Read</c> returns the same URI either way.
+        /// </remarks>
+        /// <param name="store">The parsed TriG content.</param>
+        /// <param name="graphUri">Target for triples with no graph name of their own.</param>
+        private static IEnumerable<(Uri Uri, IGraph Graph)> GroupByTargetGraph(ITripleStore store, Uri graphUri)
```
meberl: **This helper is not shared — it is copy-pasted, and the claim that it cannot drift is the reason that matters.**

Your reply said "both stores now share a `GroupByTargetGraph` helper, so the unnamed-triple routing and the per-target merge cannot drift between them." But there are two `private static` copies, one in each store file. I diffed them:

```
$ diff <(sed -n '505,537p' FusekiStore.cs) <(sed -n '490,522p' GraphDBStore.cs)
IDENTICAL - pure copy-paste
```

Byte-identical today, and free to diverge tomorrow — with the divergence showing up as triples quietly landing in the wrong graph on one backend only, which is the hardest version of this bug to notice.

This is the same shape as the `RuleSetGraphs` finding earlier in this PR, and it has the same fix. Both classes derive from `StoreBase` (`FusekiStore.cs:43`, `GraphDBStore.cs:44`), so a `protected static` member there is genuinely shared, and both call sites already pass the same two arguments. That also gives the next backend the behaviour for free instead of a third copy.

If you would rather keep the copies, the fix is to drop the "cannot drift" claim — but given the choice is available for the cost of moving one method, sharing it seems clearly better.

Trinity.GraphDB/GraphDBStore.cs
```diff
@@ -480,6 +483,43 @@ public override Uri Read(Uri graphUri, Uri url, RdfSerializationFormat format, b
...
+        /// </summary>
+        /// <remarks>
+        /// A graph carrying its own name is written under that name. Triples carrying none go to
+        /// <paramref name="graphUri"/>, the graph the caller asked to read into: they have no home of
+        /// their own, and silently discarding them would be data loss the caller cannot detect, since
+        /// <c>Read</c> returns the same URI either way.
+        /// </remarks>
+        /// <param name="store">The parsed TriG content.</param>
+        /// <param name="graphUri">Target for triples with no graph name of their own.</param>
+        private static IEnumerable<(Uri Uri, IGraph Graph)> GroupByTargetGraph(ITripleStore store, Uri graphUri)
```
meberl: The second copy. See the Fuseki side — moving this to `StoreBase` removes both.

Agreed on leaving the `ListGraphs()`/`HasGraph` inconsistency to ADR-0043's follow-up list: it is genuinely inert now that `exists` is no longer consulted, and bundling an unrelated cleanup into this PR is how the branch would stop being reviewable.

Trinity.Tests/Store/MultiGraphTrigTest.cs
```diff
@@ -0,0 +1,132 @@
...
+        /// than being dropped.
+        /// </summary>
+        /// <remarks>
+        /// The first fix for the multi-graph defect skipped every graph whose name was not an IRI,
+        /// which includes a TriG file's default graph -- so a file mixing unnamed triples with named
+        /// ones lost the unnamed ones silently, and <c>Read</c> returned the same URI either way.
+        /// nco.trig happens to name both of its graphs, so nothing in the suite would have noticed.
+        /// </remarks>
+        [Test]
+        public virtual void UnnamedTrigTriplesGoToTheGraphTheCallerNamed()
```
meberl, line 86: **The test covers the routing but not the merge — and the merge is the subtler half.**

Your reply gave a specific motivation for grouping rather than writing each graph as it comes: "two graphs can share one target (a file with unnamed triples *and* a graph named `graphUri`), and on a connector whose `SaveGraph` is a PUT the second would replace the first."

This file does not produce that collision. The named graph is `http://example.org/trig/named`, while the caller's target is `BaseUri/trig-default-graph` — two distinct targets, one graph each, so the dictionary never merges and `Merge` is never exercised. Restoring the `continue` fails this test (as you verified), but removing the grouping and writing each parsed graph directly would still pass it.

One extra case closes that, and it is a two-line change — name the inner graph the same URI the caller passes:

```csharp
var trig = $"<{subject}> <http://example.org/p> \"unnamed\" .\n"
         + $"<{target}> {{ <{subject}> <http://example.org/p2> \"named\" }}\n";

// both triples must survive: PUT semantics would leave only the second
Assert.AreEqual(2, Count(target));
```

That is the case that fails without the per-target merge, and it is exactly the scenario you described as the reason for it. Worth pinning while the reasoning is fresh — a PUT-semantics connector silently dropping half a file is precisely the failure class this PR keeps finding.

meberl (commented):
Rechecked at `5b82664`. **The unnamed-triple routing is fixed and the record is corrected.** Built clean, and every number in the updated ADR-0044 table reproduces exactly on my side:

| suite | table says | I measured |
|---|---|---|
| Virtuoso | 241 / 0 / 1 = 242 | **241 / 0 / 1 = 242** |
| GraphDB | 250 / 0 / 1 = 251 | **250 / 0 / 1 = 251** |
| Fuseki | 252 / 0 / 1 = 253 | **252 / 0 / 1 = 253** |
| in-memory (.NET 10) | 608 / 3 / 7 = 618 | **608 / 3 / 7 = 618** |
| in-memory (.NET 9) | 611 / 0 / 7 = 618 | **611 / 0 / 7 = 618** |

Five suites, every cell matching. Including the totals was the right call — that is what makes a drifted row visible without re-running anything.

The rest of the follow-ups landed well. `UnnamedTrigTriplesGoToTheGraphTheCallerNamed` fails when the `continue` is restored, which is the property that matters. ADR-0044 now records that both `nco.trig` graphs are explicitly named and that one uses a prefixed label a grep for `^<http…> {` will miss — that note is worth more than the correction itself, since it names the trap rather than just avoiding it. And swapping the `CLAUDE.md` passed-count for the quarantine count, qualified with "on net8.0", fixes the rot at its source.

Two things in `5b82664` itself, below. The first contradicts a claim in your reply, so I would not leave it as-is.

Trinity.Fuseki/FusekiStore.cs
meberl: You are right, and the claim was the worse half of it. Fixed in `37e1ddf`.

I asserted "both stores now share a `GroupByTargetGraph` helper, so … cannot drift between them." There were two `private static` copies. Byte-identical, as your diff shows, and nothing whatsoever stopping them diverging — I described the property I wanted rather than the one the code had.

Moved to `StoreBase` as `protected static`, which both classes already derive from; both call sites passed the same two arguments already, so the move was mechanical. 70 lines of duplication gone, one definition, and the next backend inherits the routing instead of copying it.

You are also right that this is the same shape as the `RuleSetGraphs` finding. Two of the seven original findings and one of these three come down to the same thing: **a rule written down twice**. Worth noting that in both cases the duplicate was byte-identical and harmless *today* — the cost is entirely in the future, which is exactly why it is easy to wave through.

Trinity.GraphDB/GraphDBStore.cs
meberl: Removed with the Fuseki copy in `37e1ddf` — one definition in `StoreBase` now, both stores call it. GraphDB measures **251 / 0 / 1 = 252** with the third TriG test.

And agreed on the `ListGraphs()`/`HasGraph` inconsistency staying on ADR-0043's list. It is inert now that `exists` is not consulted, and this branch has already grown from test infrastructure into runtime changes on two adapters; folding in an unrelated cleanup is how it would stop being reviewable at all.

Trinity.Tests/Store/MultiGraphTrigTest.cs
meberl: Correct, and this is the sharper of the two — my test proved the routing and left the justification I gave for the *merge* untested.

You are right that the two graphs in that file have distinct targets, so the dictionary never collides, `Merge` never runs, and removing the grouping entirely would still pass. I had verified the test fails when the `continue` is restored and stopped there, which only covers one of the two properties the code has.

`TrigGraphsSharingATargetAreMergedNotOverwritten` added in `37e1ddf`, with the inner graph named the same URI the caller passes — the collision I described as the reason for grouping:

```
<subject> <p1> "unnamed" .
<target>  { <subject> <p2> "named" }     <- same URI the caller passes
```

Both triples must survive; PUT semantics would leave one. Verified by defeating only the merge (unique dictionary key per parsed graph, so entries never combine): the new test fails and the other two still pass, so it isolates the property rather than just adding coverage.

The general lesson I am taking: "verified to fail when the fix is reverted" is only as good as *which* part you revert. Reverting the conspicuous half of a two-part fix proves the conspicuous half.

meberl (commented):
Rechecked at `37e1ddf`. **Both findings fixed, and I have no further ones.**

Verified rather than taken on report:

- `GroupByTargetGraph` now has exactly one definition — `StoreBase.cs:674`, `protected static` — and both stores call it. `git grep` confirms no third copy survives.
- `TrigGraphsSharingATargetAreMergedNotOverwritten` names the inner graph with the same URI the caller passes, so both triples resolve to one target and the assertion is `Count(target) == 2`. That is the collision the grouping exists for, and it is now pinned rather than argued.

All five suites re-run on this head, and every cell of the ADR-0044 table matches:

| suite | table says | I measured |
|---|---|---|
| Virtuoso | 241 / 0 / 1 = 242 | **241 / 0 / 1 = 242** |
| GraphDB | 251 / 0 / 1 = 252 | **251 / 0 / 1 = 252** |
| Fuseki | 253 / 0 / 1 = 254 | **253 / 0 / 1 = 254** |
| in-memory (.NET 10) | 608 / 3 / 7 = 618 | **608 / 3 / 7 = 618** |
| in-memory (.NET 9) | 611 / 0 / 7 = 618 | **611 / 0 / 7 = 618** |

Across three rounds this went from seven findings to zero, and the two that recurred were the same thing both times — a rule written down twice, byte-identical and harmless on the day it was written. Your framing of that in the commit message is the right one to keep.

One observation worth carrying forward rather than acting on here: the ADR table has now been updated three times in one PR and been wrong twice, each time because a commit added a test after the table was written. It is right *now* — I checked every cell. But nothing checks it, and this branch is the evidence that a hand-maintained count drifts faster than anyone expects. If these numbers are worth keeping, generating them from the `.trx` files the `stores` job already uploads would make the table as trustworthy as it currently looks. If they are not, the CI job is the better place to learn a suite went red.

Nothing here blocks the merge; from my side this is done.

PR #44, opened 2026-08-24
Finish in-memory inferencing: entail into a side graph, union it in per query

meberl:
## `inferenceEnabled: true` was silently returning non-inferred results

No exception, no warning — just the same answer the flag-off query gives. That is the failure mode
ADR-0041/0042 went to some lengths to forbid elsewhere, on the store most people meet first.

It read as supported because the wiring was there: `dotNetRDFStore` held an `RdfsReasoner`, built an
`InferencingTripleStore`, and called `AddInferenceEngine`. **None of it did anything.** Half-wired parts
are not a feature, and `doc/known-test-failures.md` had all three gaps diagnosed correctly the whole
time:

1. **The flag was never read.** `ExecuteQuery` went straight to `ToString()` → parse → `ProcessQuery`.
2. **No reasoner existed** for the way callers build the store — one was created only for a `schema=`
   connection string, while every test and the documented ADR-0011 flow seeds vocabularies with
   `store.Read` / `LoadGraphs`.
3. **The write path could not have fed it.** dotNetRDF materializes on `Add`; Trinity writes through
   `LeviathanUpdateProcessor`, which bypasses the inference engine.

## Entail into a side graph, union it in per query

```
inferenceEnabled: false  ->  FROM <g>                    (byte-identical to previous behaviour)
inferenceEnabled: true   ->  FROM <g> FROM <inferred(g)>
```

**Why not `AddInferenceEngine`, the obvious answer.** It writes entailments back into the graph they
came from, so they become visible to *every* query — and a store that does that cannot answer
`inferenceEnabled: false` correctly. It is the same all-or-nothing property that stops Fuseki honouring
the flag at all (#42). `IInferenceEngine.Apply(input, output)` is what makes a side graph possible, and
the side graph is what makes the flag per-query.

`StaticRdfsReasoner`, initialised from the graphs actually in the store rather than only from `schema=`
files. **No public API changed**: the dataset is widened on the *parsed* query via
`SparqlQuery.AddDefaultGraph`, so nothing is re-serialized.

## Results

`passed / failed / skipped = total`, the convention #43 adopted after its counts went stale twice.

| runtime | before | after |
|---|---|---|
| .NET 10 | 608 / 3 / 7 = 618 | **621 / 3 / 3 = 627** |
| .NET 9 | 611 / 0 / 7 = 618 | **624 / 0 / 3 = 627** |

The in-memory row depends on the runtime, not on this change — the assembly targets net8.0, and the
three `UriRef` failures appear only when it rolls forward onto .NET 10. CI installs net8.0 and sees
neither.

**Store suites untouched**, which is the point of the side graph: Fuseki 253 / 0 / 1, GraphDB
251 / 0 / 1, Virtuoso 241 / 0 / 1 — identical to #43.

**Seven tests come off the bench.** The four shared inferencing tests have their `Assert.Inconclusive`
overrides **deleted rather than reworded**, and the two LINQ `…WithInferencingEnabled` cases leave
quarantine, passing under both the model and model-group fixtures.

## Two things worth your attention

**Invalidation is coarse on purpose.** Any write drops every materialized graph. A SPARQL UPDATE does
not say which graph it touched, and guessing wrong serves stale entailments — a silent wrong answer,
which is the thing this PR exists to remove. Both staleness tests (recompute-after-write,
withdraw-after-delete) were verified to fail when the invalidation is removed.

**`CanSelectResourcesWithOperatorTypeOf` stays quarantined, deliberately.** It is not an inferencing
gap: it wants a base-type query to return subclasses with the flag **off**, which contradicts
`SparqlLinqExemplarTest.FiltersToExactType`. This PR settles the flag-*on* half of ADR-0037's open
decision and leaves the flag-off half as a semantics call I should not make by side effect.
`FiltersToExactType` passes **unchanged**, which is the proof that flag-off behaviour did not move.

## Also corrects the record

**ADR-0022 asserted "dotNetRDF in-memory applies an `RdfsReasoner`."** Untrue, and it is why the gap
kept being mistaken for working support — including by me, mid-review of #43, against a repository
whose own `known-test-failures.md` said plainly that the store does not implement inferencing.
ADR-0045 makes the claim true.

`DotNetRDFInferenceTest` covers what the shared tests cannot: that entailments stay *out* of
non-inferred queries, that they are recomputed after a write and withdrawn after a delete, that the
side graphs are not models, and that a query naming no graph is unaffected.

## Limits, stated rather than discovered later

- **RDFS only.** No OWL. `SkosReasoner` and `OwlReasonerWrapper` exist and could be offered later.
- **Whole-graph materialization** on first inferred query — fine for an in-memory store, not a strategy
  that would survive generalising to a large backend.
- **Entailments are computed per model graph**, so one that would follow only from two graphs *together*
  is not derived.

## Verifying locally

```bash
```

Store suites need Docker and are unaffected; they run in CI as the `stores` matrix job (#43).


meberl:
## Review history

Four rounds of review happened outside GitHub, so this records what changed and why — the code carries
the conclusions, not the reasoning that produced them.

The PR opened with a working feature and **six findings**. It closes with zero, and the shape of the
change is materially different from what was first proposed.

### Round 1 — six findings

Three were blocking, each making `inferenceEnabled: true` crash or change an answer that worked without
it:

1. **A relative `FROM` operand crashed.** The graphs to widen came from Trinity's record of the `FROM`
   operands, which is raw token text — relative when the query carries `BASE`. `1 row` with the flag
   off, `UriFormatException` with it on. Now taken from the parsed query's resolved `DefaultGraphNames`.
2. **`GRAPH` / `FROM NAMED` silently returned non-inferred results.** Entailments land in the default
   graph, so a named-graph pattern reads only asserted triples. Same data, same flag, different answer.
3. **Side graphs leaked**, **a failed materialization latched permanently**, and **a read mutated shared
   state** — three symptoms of one decision: caching entailments *inside* the store.

The third group got one fix rather than three patches: entailments are now computed **per query into a
throwaway dataset** and never enter the store. That removes all three by construction, and takes the
coarse-invalidation question with them. `RdfsInferenceCache` became `RdfsEntailment`, since it no longer
caches.

A fourth finding is the one worth remembering: `QueryWithoutADatasetIsUnaffectedByTheFlag` never had an
empty dataset — `Model.ExecuteQuery` injects a `FROM` — and asserted only `DoesNotThrow`. It would have
passed either way, **including while finding 2 was live, which it existed to catch**.

### Rounds 2–4 — the refusal, three times

Since the side-graph design cannot serve named-graph patterns faithfully, it refuses them. Getting that
refusal right took three passes, and each pass found the same class of gap:

| round | what was still reachable |
|---|---|
| 2 | walked only `ChildGraphPatterns` — missed a **subquery** and a **`FILTER EXISTS`** |
| 3 | walked patterns but not the query's other expressions — missed **`BIND`**, **projections**, **`HAVING`**, **`ORDER BY`** |
| 4 | no bypass found |

Round 3's hole produced a query that proves a triple in its `WHERE` clause and denies it in a projection
on the same line:

```sparql
SELECT ?s (EXISTS { GRAPH <m1> { ?s a <C2> } } AS ?viaGraph)
          (EXISTS {              ?s a <C2>   } AS ?viaDefault)
WHERE     {                      ?s a <C2>   }

  s=r1   viaGraph=false   viaDefault=true
```

The fix that finally held was **structural rather than enumerative**: the walker descends the pattern
kinds and expression slots it recognises and **throws on anything it does not**, so it fails toward
refusal instead of toward a silent answer. That is the discipline `OverlayQueryRewriter` arrived at for
ADR-0041 — and rounds 2 and 3 each cited that precedent while implementing only part of it.

### The recurring theme

Every round found the same thing: **a check that looked complete because it covered the conspicuous
cases.** Twice the verification was the weak point rather than the code — a test that could not reach
what it tested, and a "verified to fail when reverted" that reverted only the conspicuous half of a
two-part fix.

### What makes the refusal safe to keep

37 cases were probed in the final round: 19 refused, 18 accepted. The second set matters more. A
refuse-by-default walker is only safe if something pins the "still answers real queries" side, so the
suite asserts that ordinary SPARQL — `FILTER`, `OPTIONAL`, `UNION`, `MINUS`, subqueries, `BIND`,
`VALUES`, property paths, `NOT EXISTS`, aggregate projections, `HAVING` on an aggregate, `ORDER BY` and
`GROUP BY` expressions — is still accepted. Walking five more expression slots cost no legitimate query.

### Deliberate limits

- **`GRAPH`, `FROM NAMED` and `SERVICE` are refused** with the flag set, as is any construct the
  whitelist has not been taught — which will include answerable queries using SPARQL features added
  later. That is the intended direction to fail in, but it is a real cost.
- **RDFS only**, no OWL. **Recomputed per query**, no caching. **Entailments are per model graph**, so
  one following only from two graphs together is not derived.
- **`CanSelectResourcesWithOperatorTypeOf` stays quarantined.** It wants a base-type query to return
  subclasses with the flag *off*, contradicting `SparqlLinqExemplarTest.FiltersToExactType`. This PR
  settles the flag-*on* half of ADR-0037's open decision and leaves the other half as a semantics call.

### Final state

| suite | result |
|---|---|
| in-memory (.NET 9) | 654 / 0 / 3 = 657 |
| in-memory (.NET 10) | 651 / 3 / 3 = 657 |
| Fuseki · GraphDB · Virtuoso | 253/0/1 · 251/0/1 · 241/0/1 — unchanged |
| generator · vocabulary | 23 · 29 |

The store suites being untouched is the point of the side-graph design, not a side note. The three
in-memory failures are the pre-existing `UriRef` cases that appear only when the net8.0 assembly rolls
forward onto .NET 10; CI installs net8.0 and sees neither.

ADR-0045 records the design, the reasoning above, and the limits. ADR-0022's claim that the in-memory
store "applies an `RdfsReasoner`" is corrected — untrue until this lands, and the reason the gap was
repeatedly mistaken for working support.

PR #45, opened 2026-09-07
fix: make UriRef survive .NET 10, and make it usable as a mapped type

meberl:
## The bug

**.NET 10 added `IEquatable<Uri>` to `System.Uri`.** `EqualityComparer<T>.Default` prefers the strongly-typed interface over the `object` overload, so it routes *around* `UriRef.Equals` and compares fragment-blind — in every `HashSet<Uri>`, `Dictionary<Uri,…>`, `Contains`, `Distinct` and `GroupBy`.

`Trinity` targets netstandard2.0, so **consumers inherit this by merely running on .NET 10** — nothing has to be recompiled.

Measured on the same assembly:

| | .NET 8 | .NET 10 |
|---|---|---|
| `UriRef.Equals(object)` | False ✅ | False ✅ |
| `IEquatable<Uri>.Equals` | *absent* | **True** ❌ |
| `EqualityComparer<Uri>.Default` | False ✅ | **True** ❌ |

`doc/known-test-failures.md` called the three failing tests a test-runner artifact and "not currently reachable". Both claims were wrong; that entry is rewritten.

## Why the fix had to come in this order

The obvious advice — "use `UriRef`" — **was impossible to follow.** Three independent places reject or mis-route a `UriRef`, all for the same reason: an exact `typeof(Uri)` identity check, which a subclass fails.

| what broke | where | effect |
|---|---|---|
| `PropertyMapping<UriRef>` | `PropertyMapping.cs:205` | not in the `allowed` set → **threw in DEBUG builds** |
| serializing a `UriRef` value | `XsdTypeMapper.cs:224` | `Serializers` keyed `typeof(Uri)` → **threw "No serialiser available"** |
| lazy-load routing | `Resource.cs:477` | `!= typeof(Uri)` → `UriRef` mappings took the `ResourceCache` branch instead of direct-set |

So: make `UriRef` usable first, make it correct, then enforce it. Shipping the diagnostic first would have told people to do something that throws.

## What changed

**`UriRef`** — implements `IEquatable<Uri>`; declares `==`/`!=` (operators bind **statically**, so overriding `Equals` never covered them); `GetHashCode` mixes instead of AND-ing; `Equals` and `GetHashCode` now branch on the same operands rather than agreeing by accident; the `Uri` constructor carries `IsBlankId` across; `GetGuid` returns `UriRef`.

**`Resource.Equals`** compared with `==`, which silently affected `Class` (declares no equality of its own) and so type resolution for every query result. Now compares `OriginalString` like `Property.Equals` already did — which also makes it agree with `Resource.GetHashCode`, which has always hashed `OriginalString`. They disagreed before, on every runtime.

**Identity collections move to `UriRef`** — `ResourceCache`'s per-mapping set, the LINQ translator's rdf:type constraints (a fragment-blind `List<Uri>.Contains` dedupe **dropped a class constraint from the generated query**, loosening it), `RdfClass`/`RdfPropertyAttribute.MappedUri`, GraphDB's `ListGraphs().Contains`, Virtuoso's same-model guard. Anything keyed on `Uri.OriginalString` was left alone — it is fragment-safe already.

**Enforcement** — `TRIN007` (warning; the one diagnostic that does **not** mean "nothing was generated" — the mapping is emitted, the declared type is wrong) plus a `PropertyMapping<T>` throw that also fires in **Release**, which is where hand-written mappings the generator never sees would otherwise slip through.

**LINQ writer** — emitted IRIs via `AbsoluteUri` while the write path uses `OriginalString`: a round-trip mismatch for percent-encoding, dot-segments and host casing, and a **crash** for blank nodes, since `AbsoluteUri` throws on a relative URI. Both paths now go through `SparqlSerializer.SerializeUri`.

## Found while implementing

- **`select x.Uri` threw `Unsupported projection`.** Same defect: the case was gated on `node.Type == typeof(Uri)`, but `Resource.Uri` is declared `UriRef`, so the exact test never matched. Fixed, plus the `Uri`→`UriRef` bridge in `CoerceValue` it then needed.
- **The runtime backstop's first catch was the test model's own `PropertyMapping<Uri>`**, taking the suite from 654 passing to 659 failing. Migrating it to `UriRef` is the evidence that a `UriRef` mapping round-trips — which it could not do before this PR.

## Verification

**663 / 0 failed / 3 skipped on .NET 8, 9 and 10** (`DOTNET_ROLL_FORWARD=LatestPatch|Major|LatestMajor`). I confirmed by probe that roll-forward really executed on 10.0.11 — passing on one runtime is exactly what let this hide for a release. Generator 26, vocabulary 29.

Store suites unchanged: **Fuseki 253, GraphDB 251, Virtuoso 241, 0 failed each.** Steps 1 and 4 touch serialization, so this was not a formality.

New tests were run against a **baseline worktree at `develop`**: 4 of them fail without the fix. Two do not, and I corrected their doc comments rather than leave them claiming to reproduce a defect — `KeepsBothDeclaredTypesWhenReadingBack` is a behaviour pin (rdf:type reaches the resource by a second route), and the `Equals`/`GetHashCode` symmetry test pins a contract that previously held by short-circuit accident.

## Worth weighing in review

- **`TRIN007` is a warning but the runtime throws**, so in practice a `Uri`-typed mapping becomes a hard break on upgrade, not advisory. That follows the intended "warning + runtime throw" design, but the throw is the real gate — relevant before ElectrixOS/DevHub migrate.
- **One hazard survives and cannot be fixed.** `Uri a = someUriRef; a == b` still binds to `Uri.operator ==`. A static operator cannot be overridden. This is why ADR-0025's amendment now words "use `UriRef`" as a rule, not a preference.
- **`MappedUri` changed from `Uri` to `UriRef`** — source-compatible, binary-breaking, which 2.0 already is.


meberl:
All six addressed in `ab98cf0`. I reproduced each before fixing rather than patching from the description — the two you flagged as confirmed reproduce exactly as you described.

**`SparqlQueryTranslator.cs:2037`** — fixed. `ChainKind.Uri` now passes `node.Type` for both the source and target type, matching the sibling case. You're right that this is the same fix one line up that I failed to carry across; the comment there now covers both cases so the asymmetry can't reappear.

**`PropertyMapping.cs`** — fixed, and this was the worst of the six, because the PR *creates* it: TRIN007 and the constructor check force migration to `UriRef`, and the unmapped surface takes raw `Uri` (`IResource.AddProperty(Property, Uri)` is a public overload). `IsValueCompatible(new Uri(...))` returned `False`. Both the gate and `SetOrAddMappedValue` now widen, scalar and list, with the gate commented as having to agree with the setter.

**`SparqlQueryWriter.cs:292`** — you were right that this was worse than what it replaced, and I verified all three cases against dotNetRDF's parser:

```
PARSE_OK   :: SELECT ?p WHERE { _:b0 ?p ?o . }
PARSE_FAIL :: SELECT ?s WHERE { ?s ?p ?o . FILTER(?s = _:b0) }   :: BlankNodeWithIDToken
PARSE_FAIL :: SELECT ?s WHERE { ?s ?p ?o . FILTER(?s = <_:b0>) } :: Cannot resolve a Relative URI Reference
```

So there is no correct spelling at all — the pattern form parses and matches everything. The writer now refuses with `NotSupportedException`, the established pattern in that file. My test asserted `DoesNotThrow`, which pinned precisely the wrong behaviour; it now asserts the refusal, plus a second test pinning *why*, so the refusal doesn't get "fixed" back into a wildcard. The `OriginalString` half of the original change stands.

**`GraphDBStore.cs:525`** — fixed. It went through a local, so my `ListGraphs().Contains` grep never saw it. Reasoning from a pattern instead of reading the call sites, which is the same blind spot that produced the finding in the first place.

**`UriRef.cs:80`** — fixed, but not by reverting: `RelativeOrAbsolute` was load-bearing for carrying `IsBlankId` across, since a blank label is relative. It's now `Absolute` unless the source is a blank identifier, which restores the original `UriFormatException` at the call site and keeps the flag.

**`Resource.cs:428`** — fixed with the `ReferenceEquals` short-circuit.

**Docstring** — corrected; it now says the consumer gets the runtime refusal rather than "working code".

Re-verified: **666/0/3 on .NET 8, 9 and 10**; generator 26, vocabulary 29; **Fuseki 253, GraphDB 251, Virtuoso 241, 0 failed**. The three new regression tests fail against `4b9afac` — confirmed in a baseline worktree, not assumed.

meberl:
All three fixed in `51fa619`, each reproduced first. You're right that the bridge was applied to the add path only — I added the widening where the value goes in and never looked at where it comes out.

**`PropertyMapping.cs:454` / `:446`** — both branches now widen. The collection guard was also inverted, as you say, and that's the worse of the two: it *passes* for Uri/UriRef, so the value reached `IList.Remove`, which type-checks internally and drops the call. Add works, remove silently doesn't, `Commit()` re-persists.

Correcting the inversion is a **deliberate behaviour change** worth flagging: removing a value whose runtime type is a strict supertype of the element type — an instance that can never be in the collection — was a silent no-op and is now refused, consistent with the scalar branch and the method's own fallthrough. 749 store tests across three backends show nothing relied on the old leniency.

**`PropertyMapping.cs:305`** — this is the one that bothers me most, because it's self-inflicted: the enforcement I added for a *silent identity* bug created a *silent mapping* bug. One un-migrated mapping left every class after it (by `GetTypes()` order) unregistered, which raises no error at all, and TRIN007 being a warning meant the build still passed.

Fixed at the registry rather than the throw, since the fragility is general — any failing class did this, not just mine:

- `AddMappingClasses` attempts every class and reports all failures as one `AggregateException`. Same reasoning as the TRIN diagnostics being independent: someone migrating a large model wants the whole list from one run.
- `RegisterAssembly` marks the assembly registered only *after* the batch succeeds, so a retry isn't an early-return no-op.

I ran into this from the other side while writing the probe — a single bad class in `Trinity.Tests` took the whole suite down from `SetUpFixture`, which is the same defect viewed from the consumer end.

**Coverage**, as you asked: unit tests for both remove paths and for batch registration, plus `AddRemovePlainUriAgainstUriRefMappingTest` in the shared `ResourceMappingTest`, so the raw-Uri add/remove round trip runs against **all three stores** as well.

One test correction worth noting: my first version of the incompatible-value test passed on the baseline, because a `string` fails the inverted guard too — it only misbehaves for a supertype. Rewrote it to use that case so it demonstrates the actual inversion.

Verified: **670/0/3 on .NET 8, 9 and 10**; generator 26, vocabulary 29; **Fuseki 254, GraphDB 252, Virtuoso 242** (+1 each, the new shared test), 0 failed. All four new tests fail against `ab98cf0`, confirmed in a baseline worktree.

meberl:
You're right, and I had the inversion backwards in the part that matters. Fixed in `59dab28`.

I reproduced your case against `ab98cf0` and got exactly the message you did:

```
Expected: No Exception to be thrown
But was:  <System.Exception: Provided argument value was not of type Semiodesk.Trinity.Resource>
```

So `value.GetType().IsAssignableFrom(_genericType)` failed in **both** directions, and I described only the smaller one. Rejecting subclasses is the ordinary polymorphic case — `Person.Interests` is `List<Resource>`, so removing any `Resource` subclass through the mapped interface threw. The supertype case I did describe (silently dropped by `IList.Remove`) is the narrow half.

That reframes the change the way you put it: **a fix with a narrow tightening attached**, not a tightening I have to justify. My "749 store tests show nothing relied on the old leniency" line was arguing the weak half of the case — in the direction consumers actually hit, the old guard wasn't lenient at all.

**Coverage added.** You're right that the direction was untested: the existing `Interests` removal at `ResourceWriteSemanticsTest.cs:318` goes through `List<T>.Remove` on the collection itself and never enters `RemoveOrResetValue`. Added `RemovesACollectionValueOfASubclassThroughTheMappedInterface` — `RemoveProperty(foaf.interest, aPerson)` on the `List<Resource>` mapping, asserted through a `Commit()` and reload so it pins the store round trip too. It's in the shared `ResourceWriteSemanticsTest`, so it runs against all three backends, and it fails against `ab98cf0`.

Comments in `RemoveOrResetValue` now lead with the subclass rejection; the `UriRefTest` case is relabelled as the tightening half and cross-references the new test.

**On the scalar branch clearing regardless of the stored value** — agreed it's pre-existing at `:464`, and I mirrored it deliberately rather than diverging from it mid-fix. It's recorded in the commit message so it doesn't get lost, but it's a separate question from this PR.

Verified: **671/0/3 on .NET 8, 9 and 10**; generator 26, vocabulary 29; **Fuseki 255, GraphDB 253, Virtuoso 243**, 0 failed.

meberl:
Took the nit — it makes the test stronger than a tidy-up, so thanks for raising it.

`d8fd12d` removes from the loaded copy, using the value read back out of the mapping. Two things that buys beyond matching the consumer sequence:

- The lazy read (`ResourceCache`) now sits in front of `RemoveOrResetValue`, so the value being removed is one the mapping *resolved from the store* rather than one it still held from the `AddProperty` call. That's a link the previous version genuinely didn't cover — you were right that `loaded` was doing nothing after the precondition assertion.
- Added an `IsInstanceOf<Person>` assertion on the value read back, so the test can't quietly stop covering the subclass case it's named for if the read path ever starts handing back base `Resource` instances.

Still fails against `ab98cf0` with the same message, so the guard is unchanged in strength.

Verified: **671/0/3 on .NET 8, 9 and 10**; generator 26, vocabulary 29; **Fuseki 255, GraphDB 253, Virtuoso 243**, 0 failed. CI was green on `59dab28` across all four jobs and is re-running for this commit.

Thanks for the three rounds — the two that mattered most were both cases where my own enforcement had introduced a silent failure (the registry truncation, and the add/remove asymmetry), which is exactly the failure mode this PR set out to remove. Nothing outstanding from my side either; the scalar-branch equality question is parked in `59dab28`'s message for whoever picks it up.

PR #47, opened 2026-09-22
fix: bind bulk lazy-load subjects with VALUES, not an equality chain

meberl:
## The defect

`IModel.GetResources(IEnumerable<Uri>, Type, ITransaction)` is the bulk lazy load under **every**
mapped-property dereference (`ResourceCache.LoadCachedValues`). `Model` and `ModelGroup` built its
subject constraint as an equality chain:

```sparql
SELECT ?s ?p ?o WHERE { ?s ?p ?o. FILTER(?s = <a>||?s = <b>|| … ) }
```

Virtuoso parses that as nested binary pairs and its SPARQL compiler caps the depth. Past a threshold,
every read of a mapped collection fails with `SP031` — **and every write too**, because
`Add`/`Remove`/`Link` read the collection before mutating it. A consumer on `2.0.0-rc.3` hit this at
158 elements.

`LayeredModel` (ADR-0041) already used `VALUES` and was unaffected. Three copies of one overload is
what let this survive.

## Measured, not assumed

Through `VirtuosoStore.ExecuteQuery` — the ADO path a consumer actually hits, not the HTTP endpoint:

| subjects | equality chain | `VALUES` |
|---|---|---|
| 157 | OK, 33 ms | OK, 13 ms |
| 300 | OK, 132 ms | OK, 28 ms |
| 1000 | OK, 957 ms | OK, 83 ms |
| 2000 | **SP031** | OK, 170 ms |

Bisected thresholds, identical on **7.2.12 and 7.2.14**: the chain compiles **1024** subjects and
fails at 1025 (`SP031: sparp_gp_trav_int(): stack overflow`); `VALUES` compiles **4094** and fails
with `SP030: Too many arguments for standard built-in function`.

Two things this settled that the report could not:

- **The reported 157/158 did not reproduce.** On both these builds the chain survives 1024. It is
  also not tunable — `ThreadStackSize = 60000` moved it not at all, despite the error naming a stack
  overflow. So the threshold is a compile-time constant of someone else's server build, and *no
  chunk size for the chain is defensible*. That argues for removing the shape, not resizing it.
- **The `VALUES` ceiling is an argument count, not a text length.** 4094 subjects is 172 KB and
  compiles; 4095 does not. Chunking has to count subjects; shorter IRIs buy nothing.

After, through the shipped API: 10,000 subjects in 693 ms, no ceiling in sight.

## What changed

`LayeredModelSparql.BindSubjects` → `SparqlSerializer.GenerateSubjectBinding`, **no forwarder** (a
forwarder is how a fourth copy starts). All three models share it. Batching at **1000**, a 4x margin
under the measured ceiling, which matters because the ceiling belongs to the server, not to us.

Also fixed in the same two methods:

- Empty/null `uris` omitted the constraint on `Model`, leaving `SELECT ?s ?p ?o WHERE { ?s ?p ?o. }`
  — **the whole model**. `ModelGroup` had no guard at all: empty emitted `FILTER ( )`, null threw.
- Subjects were interpolated via `Uri.ToString()`, which **unescapes percent-encoding** — an escaped
  IRI was silently not found.
- Blank ids were emitted as `<_:0>`, an invalid relative IRI that took the whole query down. No
  SPARQL query can address a blank node by label, so they are skipped and `ResourceCache`
  materializes them as unresolved.
- `uris` was enumerated twice — and the sole caller passes its **live** cache set and mutates it
  while consuming the result, so it is now copied eagerly. Without that, batch 2 throws.
- A dangling `if` guarded only `r.IsNew`. Removed on evidence: git shows it was made vestigial in
  `5b162fa` when `result.GetResources()` gained its type argument, and `GetMatchingTypes` filters by
  `type.IsAssignableFrom`, so the condition is unconditionally true.
- Added the null-resource guard `Model` lacked but `ModelGroup` had.

## The blank-node quarantine: half fixed, and I'm saying which half

The hypothesis in the brief was right about the crash and incomplete about the cure — `SerializeUri`
alone emits a bare `_:b0`, which is not a legal `VALUES` operand either. Skipping them is the fix.

Un-quarantining `CanRemoveBlankNodeValuedLink` then **fails**, one layer further in:

```
SparqlUpdateException: Cannot create a DELETE command where any of the Triple Patterns are not
constructable triple patterns (Blank Node Variables are not permitted)
```

This is exactly what `doc/known-test-failures.md` predicted would happen. The **read** half is fixed
and split out as `CanReadBlankNodeValuedLink`, which passes; the **write** half stays quarantined
under its own, now accurate, reason. `known-test-failures.md` and ADR-0039 updated to match.

## Tests — and why 300 is not the guard

A 300-member round-trip **does not catch this bug** on a stock Virtuoso, and I verified that rather
than assuming it: with the fix reverted on purpose, the 300-member test still passed and only the
2000-subject test failed, with `SP031`, on both fixtures. The guard is therefore
`GetResourcesByUriCompilesBeyondTheEqualityChainLimit`, which asks for 2000 subjects that need not
exist — the failure was in *compiling* the query, so it costs one query and no writes.

Added to the abstract `ModelTest<T>` / `ModelGroupTest<T>`, so all four fixtures inherit them
(in-memory + Virtuoso + GraphDB + Fuseki): 300-member round-trip incl. a write, empty/null contracts,
blank-node member, and the above-the-limit guard. Plus unit tests pinning that
`ProvidesStatements()` survives the new shape — **in both directions**, including that dropping the
trailing `.` would silently turn materialization off.

## Verification

| suite | result |
|---|---|
| in-memory (.NET 8, 9, 10) | 751 / 0 / 3 |
| Virtuoso | 312 / 0 / 1 (was 303 / 0 / 1) |
| GraphDB | 323 / 0 / 1 (was 314 / 0 / 1) |
| Fuseki | 325 / 0 / 1 (was 316 / 0 / 1) |
| generator | 26 / 0 / 0 |
| vocabulary | 29 / 0 / 0 |

## Deliverables

- `doc/adr/0046-bulk-subject-binding-with-values.md`, indexed in the ADR README
- `CHANGELOG.md` — new file; `RELEASING.md` gate 7 referenced one that never existed
- `PackageReleaseNotes` in `Directory.Build.props` (renders on the nuget.org listing)
- `CLAUDE.md` invariant + refreshed counts

**Not packed or published** — this is code and docs only, per the agreed scope. `2.0.0-rc.4` is a
`-p:VersionSuffix=rc.4` at pack time; nothing in the repo pins the label.


meberl:
-

meberl:
-

meberl:
-

meberl:
-

meberl:
-

meberl:
-

meberl:
-

meberl:
-

